PR #8, opened 2026-04-09
Fix misc. metalman bootstrapping issues

jveski:
- Adds `--dhcp-auto-interface` to the metalman controller deployment
- Structured cloudinit logs instead of raw json
- Add missing UEFI boot entry to host image
- Rename smoke.py to smoke-metalman.py
- Move metalman RBAC manifest so it actually gets deployed
- Discover the apiserver url from kube-public instead of trusting the in-cluster config (which typically uses apiserver's clusterip)


jwilder (approved):

jveski:
I'm going to merge these fixes for now and will fix the flaky smoke tests later. It's more important to get the fixes merged now to avoid more merge conflicts.

PR #37, opened 2026-04-14
[WIP] Machine CR watch-based daemon

bcho:
## Summary

Implemented a Kubernetes watch-based daemon that monitors the Machine CR on the control plane and reconciles the local node to match the desired state.

### What changed

**Daemon** (`cmd/agent/internal/daemon/`)
- Watch-based loop using `client.NewWithWatch` on the Machine CR matched by `MachineName` from the applied config
- Async reconciliation via client-go workqueue (decouples watch events from reconciliation)
- Authenticates via bootstrap token from the applied config (`system:bootstrappers` group) - avoids reading kubeconfig files from inside the nspawn machine (which contain nspawn-internal paths). A TODO marks this as temporary pending a proper agent credential strategy.
- Registers a Machine CR at startup if one does not already exist (supports dynamic/manual-bootstrap environments)
- Detects operation counter drift (`spec.operations.repaveCounter > status.operations.repaveCounter`) to trigger reconciliation. Reboot counter is not handled by the daemon.
- Inside `updateNode`, checks actual config drift (version, image) before performing the expensive rootfs reprovision
- On success: sets phase to `Joining`, acknowledges repave counter (spec -> status), sets `NodeUpdated` condition to `True/Succeeded`
- On failure: sets phase to `Failed`, sets `NodeUpdated` condition to `False/Failed`
- Before reconciliation: sets phase to `Provisioning`, sets `NodeUpdated` condition to `False/InProgress`
- Re-GETs Machine CR from API server at start of each event handler to avoid stale watch events causing spurious drift
- Kube client built via parameter injection (unexported `kubeClientFunc` type) for testability

**Applied config integrity** (`cmd/agent/internal/goalstates/checksum.go`)
- SHA-256 sidecar checksum file written alongside applied config JSON
- Verified on read; missing sidecar = warning (proceed), mismatch = error

**Phase helpers** (extracted from inline code)
- `rootfs.Provision(log, gs)` - composite rootfs provisioning task
- `nodestart.StartNode(log, gs, cfg)` - composite node start task
- `nodestop.StopNode(log, machineName)` - composite node stop task
- `utilexec.MachineRun` - nspawn machine command execution

**Goal state** (`cmd/agent/internal/goalstates/`)
- `ResolveMachine()` discovers active nspawn machine and loads applied config
- `MachineGoalState` provides machine name, config, and rootfs paths
- `DaemonUnit` const consolidates the systemd unit name in one place

**RBAC** (`deploy/machina/07-bootstrapper-rbac.yaml`)
- Agent daemon RBAC consolidated into the bootstrapper ClusterRole (`unbounded-bootstrapper-machine`)
- Grants create/get/list/watch on `machines`, get/update/patch on `machines/status`
- Bound to `system:bootstrappers` group

**Systemd** (`unbounded-agent-daemon.service`)
- `After=network-online.target machines.target` ensures daemon starts after nspawn machines are available
- Enabled in the start command's serial task list after initial provisioning
- Stopped, disabled, and removed during reset (Step 1 before stopping nspawn machines)

**Config directories**
- Unified under `/etc/unbounded/kube` and `/etc/unbounded/agent` (was `/etc/unbounded-kube` and `/etc/unbounded-agent`)

**Design doc** (`docs/content/reference/agent/daemon.md`)
- Architecture SVG diagram, auth model, RBAC, systemd lifecycle, reconciliation flow

### E2E test flow (single linear sequence)

1. Start node without pre-created Machine CR
2. Wait for node Ready
3. Validate Machine CR (daemon self-registers it, polling up to 120s)
4. Validate daemon is watching
5. Trigger upgrade: patch Machine CR with new kube version + bump repaveCounter
6. Validate upgrade completed (daemon performs alternating nspawn machine update)
7. Wait for node Ready post-upgrade
8. Validate workload scheduling
9. Reset agent + delete Machine CR
10. Rejoin (run agent again on same VM, no VM recreation)
11. Wait for node Ready
12. Validate daemon watching
13. Validate workload

### Removed

- `nodestart.RegisterMachine` phase (moved to daemon)
- `deploy/agent/01-rbac.yaml` (consolidated into bootstrapper RBAC)
- Two-case e2e structure with VM recreation between cases

### DIFF
--- a/cmd/agent/internal/daemon/daemon.go
+++ b/cmd/agent/internal/daemon/daemon.go
@@ -9,27 +9,33 @@ import (
 	"fmt"
 	"log/slog"
 	"strings"
+	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	v1alpha3 "github.com/Azure/unbounded-kube/api/machina/v1alpha3"
 	"github.com/Azure/unbounded-kube/internal/provision"
 )
 
-// kubeClientFunc constructs a controller-runtime client from a rest.Config.
-// The production implementation is client.NewWithWatch; tests can supply a fake.
+// kubeClientFunc constructs a controller-runtime WithWatch client from a
+// rest.Config. The production implementation is client.NewWithWatch;
+// tests can supply a fake.
 type kubeClientFunc func(cfg *rest.Config, opts client.Options) (client.WithWatch, error)
 
+const (
+	// watchRetryInterval is the delay between watch re-establishment attempts.
+	watchRetryInterval = 10 * time.Second
+)
+
 // Run is the main daemon entry point. It discovers the active nspawn
-// machine, builds a Kubernetes client, registers the Machine CR if needed,
-// and blocks until the context is cancelled.
-//
-// TODO: Add a trigger mechanism (e.g. file watch, signal, API) to invoke
-// updateNode when the desired config changes.
+// machine, builds a Kubernetes client from the applied config, and
+// watches the Machine CR for changes.
 func Run(ctx context.Context, log *slog.Logger) error {
 	return run(ctx, log, client.NewWithWatch)
 }
@@ -43,13 +49,14 @@ func run(ctx context.Context, log *slog.Logger, newClient kubeClientFunc) error 
 		return fmt.Errorf("find active machine: %w", err)
 	}
 
+	machineName := active.Config.MachineName
 	log.Info("daemon starting",
-		"machine_cr", active.Config.MachineName,
+		"machine_cr", machineName,
 		"nspawn_machine", active.Name,
 		"applied_version", active.Config.Cluster.Version,
 	)
 
-	// Build a controller-runtime client from the applied config.
+	// Build a controller-runtime WithWatch client from the applied config.
 	kubeClient, err := buildKubeClient(active.Config, newClient)
 	if err != nil {
 		return fmt.Errorf("build kube client: %w", err)
@@ -59,18 +66,63 @@ func run(ctx context.Context, log *slog.Logger, newClient kubeClientFunc) error 
 		"api_server", active.Config.Kubelet.ApiServer,
 	)
 
-	// Ensure a Machine CR exists before blocking. In dynamic environments
-	// (manual-bootstrap, cloud-init) a Machine CR may not have been
-	// pre-created by machina.
+	// Ensure a Machine CR exists before entering the watch loop. In
+	// dynamic environments (manual-bootstrap, cloud-init) a Machine CR
+	// may not have been pre-created by machina.
 	if err := registerMachine(ctx, log, kubeClient, active.Config); err != nil {
 		return fmt.Errorf("register machine: %w", err)
 	}
 
-	// Block until shutdown.
-	<-ctx.Done()
-	log.Info("daemon shutting down")
+	// Use client-go's rate-limiting workqueue to decouple the watch loop
+	// from reconciliation. The watch loop adds the machine name on every
+	// relevant event; the queue deduplicates and rate-limits so the
+	// worker processes at most one reconciliation at a time. This keeps
+	// the watch loop free to drain events without backpressuring the API
+	// server's HTTP/2 stream.
+	queue := workqueue.NewTypedRateLimitingQueueWithConfig(
+		workqueue.DefaultTypedControllerRateLimiter[string](),
+		workqueue.TypedRateLimitingQueueConfig[string]{Name: "machine"},
+	)
+	defer queue.ShutDown()
 
-	return nil
+	// Worker goroutine: processes items from the queue.
+	go runWorker(ctx, log, kubeClient, queue)
+
+	// Enter the watch loop.
+	for {
+		if err := watchMachine(ctx, log, kubeClient, machineName, queue); err != nil {
+			if ctx.Err() != nil {
+				return nil // Graceful shutdown.
+			}
+
+			log.Error("watch failed, retrying", "error", err, "retry_in", watchRetryInterval)
+
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(watchRetryInterval):
+			}
+		}
+	}
+}
+
+// runWorker drains the workqueue, processing one item at a time.
+func runWorker(ctx context.Context, log *slog.Logger, c client.WithWatch, queue workqueue.TypedRateLimitingInterface[string]) {
+	for {
+		key, shutdown := queue.Get()
+		if shutdown {
+			return
+		}
+
+		if err := handleMachineEvent(ctx, log, c, key); err != nil {
+			log.Error("reconciliation failed, requeuing", "error", err)
+			queue.AddRateLimited(key)
+		} else {
+			queue.Forget(key)
+		}
+
+		queue.Done(key)
+	}
 }
 
 // buildKubeClient creates a controller-runtime WithWatch client from the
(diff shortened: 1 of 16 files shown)


cmd/agent/internal/goalstates/constants.go
```diff
@@ -3,15 +3,21 @@
...
 
+import "fmt"
+
 const (
 	// ConfigDir is the host-level configuration directory for unbounded-kube.
 	ConfigDir = "/etc/unbounded-kube"
 
+	// AgentConfigDir is the host-level configuration directory for the
+	// unbounded-agent. Applied config files are persisted here.
+	AgentConfigDir = "/etc/unbounded-agent"
```
plombardi89: Can we just unify everything under `/etc/unbounded`? 

`/etc/unbounded/kube`
`/etc/unbounded/agent`

cmd/agent/internal/goalstates/resolve.go
```diff
@@ -0,0 +1,147 @@
...
+	NodeStart *NodeStart
+}
+
+// ResolveMachine probes the host (kernel version, hostname, GPU hardware) and
+// resolves the complete goal state for the named nspawn machine from an agent
+// config.
+func ResolveMachine(log *slog.Logger, cfg *provision.AgentConfig, machineName string) (*MachineGoalState, error) {
+	kernel, err := hostKernel()
+	if err != nil {
+		return nil, err
```
plombardi89: annotate error

cmd/agent/internal/goalstates/resolve.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	hostname, err := os.Hostname()
+	if err != nil {
+		return nil, fmt.Errorf("get host hostname: %w", err)
+	}
+
+	nvidia, err := ResolveNvidiaHost(runtime.GOARCH)
+	if err != nil {
+		return nil, err
```
plombardi89: annotate error

cmd/agent/internal/goalstates/resolve.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	ociImage := ResolveOCIImage(log, cfg.OCIImage, len(nvidia.GPUDevicePaths) > 0)
+
+	kubelet, err := resolveKubelet(cfg)
+	if err != nil {
+		return nil, err
```
plombardi89: annotate error

plombardi89 (changes requested):
Mostly looks good. I'd make sure all the errors are annotated as it helps debugging when something goes wrong if you have a bunch of traceable errors.

cmd/agent/internal/daemon/update.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return fmt.Errorf("resolve machine goal state: %w", err)
+	}
+
+	err = phases.Serial(log,
+		rootfs.Provision(log, gs.RootFS),
+		nodestop.StopNode(log, oldMachine),
+		nodestart.StartNode(log, gs.NodeStart, newCfg),
+		nodestart.WaitForKubelet(log, newMachine),
+		reset.CleanupMachine(log, oldMachine),
```
jveski, line 153: Shouldn't we stop the previous kubelet before starting the new one?

cmd/agent/internal/daemon/daemon.go
```diff
@@ -0,0 +1,482 @@
...
+				continue
+			}
+
+			log.Debug("watch event",
+				"type", event.Type,
+				"generation", machine.Generation,
+				"version", machine.ResourceVersion,
+			)
+
+			if err := handleMachineEvent(ctx, log, wc, machine); err != nil {
```
jveski: This should be async to avoid blocking the watch loop. Otherwise it's possible to put backpressure on apiserver when the watch and http2 buffers fill up.

You could pull in the standard `workqueue` but a `make(chan struct{}, 1)` worker would do in this case.

cmd/agent/internal/daemon/update.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// findActiveMachine scans the agent config directory for an applied config
+// file and returns the active machine name and config. Returns an error if
+// no applied config is found.
+func findActiveMachine() (*ActiveMachine, error) {
+	for _, name := range []string{goalstates.NSpawnMachineKube1, goalstates.NSpawnMachineKube2} {
+		path := goalstates.AppliedConfigPath(name)
+
+		data, err := os.ReadFile(path)
```
jveski, line 51: If we are going to consider this file authoritative it deserves a checksum to protect against bitflips

cmd/agent/internal/daemon/daemon.go
```diff
@@ -0,0 +1,482 @@
...
+			machine.Status.Conditions[i].Message = message
+			machine.Status.Conditions[i].LastTransitionTime = now
+			machine.Status.Conditions[i].ObservedGeneration = machine.Generation
+			found = true
+
+			break
+		}
+	}
+
+	if !found {
```
jveski: nit: controller-runtime has a nice helper for ensuring a condition exists

cmd/agent/internal/daemon/daemon.go
```diff
@@ -0,0 +1,482 @@
...
+		machine.Status.Conditions = append(machine.Status.Conditions, metav1.Condition{
+			Type:               "NodeUpdated",
```
jveski: nit: use a const in the api package to document the new condition

cmd/agent/internal/daemon/daemon.go
```diff
@@ -0,0 +1,482 @@
...
+		return nil
+	}
+
+	log.Info("operation counter drift detected",
+		"current_version", active.Config.Cluster.Version,
+		"desired_version", specVersion(machine),
+	)
+
+	// Set status to Provisioning before starting work.
+	if err := updateMachinePhase(ctx, c, machine, v1alpha3.MachinePhaseProvisioning, "agent daemon reconciling"); err != nil {
```
jveski, line 323: Also need to set a condition here. The phase should just follow conditions without being authoritative

cmd/agent/internal/daemon/update.go
bcho: nodestop.StopNode(log, oldMachine) is for stopping the previous one

cmd/agent/internal/daemon/daemon.go
```diff
@@ -0,0 +1,482 @@
...
+	if cfg.Cluster.CaCertBase64 == "" {
+		return nil, fmt.Errorf("applied config has no CA certificate")
+	}
+
+	caData, err := base64.StdEncoding.DecodeString(cfg.Cluster.CaCertBase64)
+	if err != nil {
+		return nil, fmt.Errorf("decode CA certificate: %w", err)
+	}
+
+	if cfg.Kubelet.BootstrapToken == "" {
```
jveski, line 148: The bootstrap tokens won't live forever. Maybe it can wait for kubelet to go through its bootstrap workflow and then just use the kubelet's kubeconfig. But I think it's probably better to sign our own certs for the agents so they'll (hopefully) still work when kubelet is broken.

bcho: yeah this part needs discussion: we need to define what cred to use for this agent to talk to the kube-apiserver. I don't have a clear answer for this part yet

cmd/agent/internal/goalstates/constants.go
bcho: it turns out this might need a bigger change, i will do it in a separate PR

cmd/agent/internal/daemon/daemon.go
bcho: switched to use workqueue for this

cmd/agent/internal/daemon/daemon.go
bcho: added condition inside upateMachinePhase call

cmd/agent/internal/goalstates/constants.go
bcho: refactort that part as well in this pr

cmd/agent/internal/daemon/update.go
bcho: added support for generating a checksum sidecar file below

cmd/agent/internal/daemon/daemon.go
bcho: added a comment below to call out this is a temporary solution, will follow up in future pass

PR #46, opened 2026-04-16
feat: new CRs preliminary doc

phealy:
Comment and feedback on new CR proposal.


docs/content/concepts/new_crds.md
```diff
@@ -0,0 +1,60 @@
...
+
+When a user creates a `MachineConfiguration` object, a corresponding `MachineConfigurationVersion` child object (using a Kubernetes `ownerReference`) named `<MachineConfigurationName>-v<version>` and a `status.deployed` of `false` will be automatically created by the controller. This version will remain editable until it has been deployed (see the `Machine` object). Once any machines have been deployed using a given configuration version, the state of the configuration will be changed to deployed and the spec fields of the object will become immutable like a running `Pod` (this should be enforced by a `ValidatingAdmissionPolicy`). The controller may still update the status field to include things like the count of machines currently referencing the policy. Once deployed, a version may never be modified again, but may be deleted if it is not currently referenced. Deletion of a version that is still referenced by any `Machine` object should be blocked by a `ValidatingAdmissionPolicy`. The highest version created should be stored in a status field on the `MachineConfiguration` so that the controller will not reuse a number if a `MachineConfigurationVersion` is deleted.
```
jwilder: What happens if you edit a MachineConfiguration?  Is that blocked?  Or would editing a MachineConfiguration trigger the creation of a new MachineConfiguationVersion?

Just thinking that it may be more common for someone to edit a MachineConfiguration vs copying and creating a new derivative.  For example, updating to a new OS version would just require editing one field.

docs/content/concepts/new_crds.md
```diff
@@ -0,0 +1,60 @@
...
+
+When changes are made to a `MachineConfiguration` object specification, the controller will copy the changes to the most recent `MachineConfigurationVersion` if it has not been deployed or create a new version if there are no non-deployed versions. In this way, the count of versioned objects does not grow extremely large, as they only become locked and permanent when deployed, and old configurations may be deleted.
```
jwilder, line 35: Oh.  So updates continue to mutate the MachineConfigurationVersion.  So we effectively have a 1:1 relationship between a MachineConfiguration and a MachineConfigurationVersion.  I was thinking the MachineConfigurationVersion was more like an history of changes to a MachineConfiguration, but we dont' get that with the "most recent" concept. 

If we don't keep a history, could we just use the MachineConfiguration with a "deployed" indicator?  If it's deployed, edits are blocked by the validation policy.

jwilder (commented):
A few questions.  Makes sense though.

docs/content/concepts/new_crds.md
```diff
@@ -0,0 +1,60 @@
...
+
+The configuration selected drives the VM agent (see Agents, below) to create the `systemd-nspawn` container and bootstrap `kubelet` such that the node joins the cluster. Once provisioning has completed and the node has appeared in the cluster, the machine controller should annotate the `Node` object with the `MachineConfiguration` name and version that were used to bootstrap the machine.
+
+If the user wants to repave the `systemd-nspawn` container or make configuration changes on the VM effective, they should cordon/drain the node and then delete the Kubernetes `Node` object. The VM agent will detect that the `Machine` exists but no matching Kubernetes `Node` exists; this will then trigger the repave operation on the node, which should result in the node rejoining the cluster and becoming ready.
```
jveski, line 19: Are we confident we'll never need the MX/CX separation? i.e. Machines on one cluster and Nodes on another(s)? @jwilder

docs/content/concepts/new_crds.md
phealy: When a MachineConfiguration is edited:
- If an editable/non-deployed MachineConfigurationVersion exists for that MachineConfiguration, it is updated.
- Else: a new editable MachineConfigurationVersion is created with the next available version number.

So in other words:
1. I first create MachineConfiguration `myconfig`. MachineConfigurationVersion `myconfig-1` is created after the object is validated and created.
2. I make some changes to the profile as I read the docs. `myconfig-1` is still editable, so it's just updated.
3. I deploy a machine from `myconfig-1`. `myconfig-1` is now immutable (except for status). The controller updates the `status.deployedMachines` to `1`. The node object that joined the cluster is annotated with `unbounded/MachineConfigurationVersion: myconfig-1`.
4. I make changes to `myconfig`. Since `myconfig-1` is immutable, `myconfig-2` is automatically created. Any future changes are applied to `myconfig-2` until it's deployed. `myconfig-1` cannot be deleted because a ValidatingAdmissionPolicy rejects deletes on MachineConfigurationVersions with `status.deployedMachines != 0`.

Does that clarify things? Your actual MachineConfiguration is always editable, like a Deployment is. If you make a change to it, it creates or updates a MCV, like a ReplicaSet.

docs/content/concepts/new_crds.md
phealy: See my response to the earlier comment - we *do* get history, because we only mutate the MCV until it's actually been used. As soon as an MCV is deployed anywhere, it's now immutable and gives us that history. This is to avoid having a bunch of small changes to a MachineConfiguration that are never deployed triggering a bunch of useless MCVs in the data store - my logic was that if it was never actually used, we probably don't care about that version.

docs/content/concepts/new_crds.md
jveski: > this should be enforced by a ValidatingAdmissionPolicy

nit: immutability can be enforced with cel expressions on the crd itself

jveski: > but may be deleted if it is not currently referenced

Add something like a `revisionhistorylimit`?

docs/content/concepts/new_crds.md
```diff
@@ -0,0 +1,60 @@
...
+
+The `MachineOperation` will take a spec field that references a `Machine` object (or a `machineSelector` to target multiple machines, in which case the controller creates individual `MachineOperation` children per matched machine) and an `operationName` field that's an enum-as-string, so it can contain both predefined operations (like `reboot`, `shutdown`, `restartService`, `powerCycle`, `powerOff`, or `powerOn`) or custom operations that are supported by individual cloud controllers (these might be troubleshooting actions like `packetCapture` that could be performed on any machine or cloud-specific actions that may be needed). Different agents may act on the operations based on the name - for example, the in-VM agent would handle operations like `restart` and `shutdown`, which would be operating system level, whereas a cloud or PXE agent would handle things like `powerOff` that need to be handled outside of the VM.
```
jveski: > Different agents may act on the operations based on the name - for example, the in-VM agent would handle operations like `restart` and `shutdown`, which would be operating system level, whereas a cloud or PXE agent would handle things like `powerOff` that need to be handled outside of the VM.

It would be worth designing this such that agents only need to watch their own operations. My first thought is using a controller to "denormalize" an active operation into the machine status. Or maybe label operations by machine name and use a label selector to scope down the agent's informer.

docs/content/concepts/new_crds.md
```diff
@@ -0,0 +1,60 @@
...
+
+A `parameters` field of type `map[string]string` should be supplied that can be used to pass parameters to operations if required.
```
jveski, line 49: Any reason not to use annotations for these values? Hopefully most operations won't need to use it anyway.

docs/content/concepts/new_crds.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Agents
```
jveski, line 51: It would be helpful to describe each responsibility in terms of the current components e.g. machina, metalman, agent.

docs/content/concepts/new_crds.md
jveski: What kind of guarantees will this api provide re. ordering or machine writes? For example: if I update a machine to a new config, then immediately create an operation (let's say pxe reimage) am I guaranteed that the operation will "see" the latest configuration?

Or maybe the operation status captures the generation of the machine at the time that was used by the operation. Then the client can always send another operation if needed.

docs/content/concepts/new_crds.md
```diff
@@ -0,0 +1,60 @@
...
+
+Machines have a 1:1 relationship with a Kubernetes `Node` object. When a machine is initially added to the cluster, it should trigger initial provisioning of the VM or compute, but *not* add the node to the cluster. That should be done only after a `MachineConfiguration` has been selected, which may happen either by setting the `MachineConfiguration` (and possibly version) manually on the `Machine` object or via the `MachineConfiguration` CR having a `machineSelector` that matches fields on the `Machine` object. If no `MachineConfiguration` has been selected and no `machineSelector` matches, the `Machine` should report a `ConfigurationPending` status condition and remain in a waiting state until a configuration is assigned.
```
plombardi89, line 15: I'm not so sure about this 1:1 relationship. It locks us out of some useful scenarios where you want to partition a large piece of hardware into many machines. Think a use-case like extending workload runners into a DC site but the workloads do not each need 100% of the machine's capability, for example, CI workloads.

docs/content/concepts/new_crds.md
jwilder: Yes, I think the mx/cx distinction is not relevant anymore.

docs/content/concepts/new_crds.md
phealy: Are you thinking we'd have multiple nspawn containers on the same node?

My initial thought on that is 1:1 `Machine`:`Node`, but we could add a `MachineHost` that is effectively a virtualization host which can be carved up into many `Machines`.

jwilder: I wonder if that use case might be better served by running VMs on the machines and the VMs can join as nodes too?  Or would just regular pods w/ requests/limits be sufficient?

docs/content/concepts/new_crds.md
phealy: TIL! It doesn't look like that can be done for deletes, though, so we'd still need a VAP for delete blocking, I think - do you concur? See the comment below for modified text.

phealy: We'd have to be explicit that that only applies to versions that don't have any currently deployed machines - wouldn't want to delete a version that is still in use.

```suggestion
When a user creates a `MachineConfiguration` object, a corresponding `MachineConfigurationVersion` child object (using a Kubernetes `ownerReference`) named `<MachineConfigurationName>-v<version>` and a `status.deployed` of `false` will be automatically created by the controller. This version will remain editable until it has been deployed (see the `Machine` object). Once any machines have been deployed using a given configuration version, the state of the configuration will be changed to deployed and the spec fields of the object will become immutable like a running `Pod` (this should be enforced by the CRD schema using CEL/`x-kubernetes-validations`). The controller may still update the status field to include things like the count of machines currently referencing the policy. Once deployed, a version may never be modified again, but may be deleted if it is not currently referenced. Deletion of a version that is still referenced by any `Machine` object should be blocked by a `ValidatingAdmissionPolicy`. The highest version created should be stored in a status field on the `MachineConfiguration` so that the controller will not reuse a number if a `MachineConfigurationVersion` is deleted. An integer `revisionHistoryLimit` on the MachineConfiguration will trigger the controller to only retain the last `<limit>` versions, though this will not override the restriction on deleting in use configurations, so more configurations than the limit may be retained.
```

docs/content/concepts/new_crds.md
plombardi89: > I wonder if that use case might be better served by running VMs on the machines and the VMs can join as nodes too? Or would just regular pods w/ requests/limits be sufficient?

When we decided on the systemd-nspawn approach I always was just considering it as an impl detail under the hood to solve the host-layer system pollution and reset problem. The agent could spin up kvm based nodes as well, might be a little more complex with the networking, but would provide strong isolation which is probably necessary for these shared workload scenarios. This should be configurable somehow even if we don't plan to implement that kvm-node (or other alt virtualization tech) approach anytime soon.

docs/content/concepts/new_crds.md
phealy:
```suggestion
The `MachineOperation` will take a spec field that references a `Machine` object (or a `machineSelector` to target multiple machines, in which case the controller creates individual `MachineOperation` children per matched machine) and an `operationName` field that's an enum-as-string, so it can contain both predefined operations (like `reboot`, `shutdown`, `restartService`, `powerCycle`, `powerOff`, or `powerOn`) or custom operations that are supported by individual cloud controllers (these might be troubleshooting actions like `packetCapture` that could be performed on any machine or cloud-specific actions that may be needed). Different agents may act on the operations based on the name - for example, the in-VM agent would handle operations like `restart` and `shutdown`, which would be operating system level, whereas a cloud or PXE agent would handle things like `powerOff` that need to be handled outside of the VM. The controller should label the `MachineOperation` with the `Machine` name so that the agents can use a `labelSelector` on their informers.
```

docs/content/concepts/new_crds.md
phealy: Is that cleaner than having a parameters array? I lean towards not putting stuff like that in the metadata, since it's really part of the spec, but I'm definitely open to annotations if we'd prefer that.

docs/content/concepts/new_crds.md
phealy: I'm less up to date with what machina/metalman do - would you mind suggesting what you'd propose?

docs/content/concepts/new_crds.md
phealy: Well, having separate `systemd-nspawn` containers provides more isolation than separate pods and less than VMs, but having a `MachineHost` that can be of type `kvm` or `nspawn` or whatever else we want would cover multiple cases. It could even be extended to cover KubeVirt - a `MachineHost` with `type: KubeVirt` could then be used to create/define multiple child machines that would then be joined back as nodes to the same cluster.

docs/content/concepts/new_crds.md
phealy: The suggestion seems to meet the need - easy enough to capture the referenced `MachineConfigurationVersion` in a status field for reimage operations. Is it necessary, though, given that the node object should be annotated with the configuration once it's applied and the `Machine` should also have the same information in the status fields?

docs/content/concepts/new_crds.md
jwilder: This seems like we're mixing workload with host management concerns.  Machine->Node is about managing the core compute.  Running workloads (like runners) as pods/vms seems orthogonal to that.  You could run the in vms w/ kube-virt or pods and carve up the host how you want.  It's not dependent on how the node is joined underneath.  

I do like the idea of being flexible in how the agent runs though (nspawn, on host, in vm, etc..).  Not sure if that adds unnecessary complexity though.

phealy: I don't mean running workloads like runners or pods, I mean a case of dividing a large compute node up into several smaller, logical compute nodes that all register to kubernetes and are managed by unbounded.

docs/content/concepts/new_crds.md
```diff
@@ -0,0 +1,60 @@
...
+
+The configuration selected drives the VM agent (see Agents, below) to create the `systemd-nspawn` container and bootstrap `kubelet` such that the node joins the cluster. Once provisioning has completed and the node has appeared in the cluster, the machine controller should annotate the `Node` object with the `MachineConfiguration` name and version that were used to bootstrap the machine.
+
+If the user wants to repave the `systemd-nspawn` container or make configuration changes on the VM effective, they should cordon/drain the node and then delete the Kubernetes `Node` object. The VM agent will detect that the `Machine` exists but no matching Kubernetes `Node` exists; this will then trigger the repave operation on the node, which should result in the node rejoining the cluster and becoming ready.
+
+**TODO**: Figure out a way to deal with changes that don't require a repave, if we want to allow that. Perhaps an `applyConfiguration` operation that can be used to indicate we want changes applied, which could take a parameter for disruptive or not, and handle the automation of cordon/drain/delete?
```
bcho, line 21: we can potentially put one or more hash annotations to the object to indicate the applied settings version? Similar approach use in karpenter side: https://karpenter.sh/docs/concepts/nodeclaims/#:~:text=aws/tagged%3A%20true-,karpenter.sh/nodepool%2Dhash,-%3A%20377058807571762610%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20karpenter.sh

bcho: but do we want to detect this in controller (outside of the node) or through agent (inside the node)?

docs/content/concepts/new_crds.md
bcho: reusing your example, after repaving an machine X from `myconfig-1` to `myconfig-2`, what will be the deployedMachines / deployed status values for `myconfig-1` ? Do we still allow editing in this case?

Mostly the question is around how to determine the deployed status and if it can be changed after machines moving away

docs/content/concepts/new_crds.md
```diff
@@ -0,0 +1,60 @@
...
+
+When changes are made to a `MachineConfiguration` object specification, the controller will copy the changes to the most recent `MachineConfigurationVersion` if it has not been deployed or create a new version if there are no non-deployed versions. In this way, the count of versioned objects does not grow extremely large, as they only become locked and permanent when deployed, and old configurations may be deleted.
+
+#### Rollback
+
+To roll back to a previous configuration, the user updates the `Machine` object's configuration version reference to point to an earlier `MachineConfigurationVersion`. The normal repave workflow (cordon/drain/delete the `Node`) then applies the prior version. This avoids the need for a dedicated rollback mechanism — any previously deployed version can be targeted directly.
+
+## Operational CRs
+
+### MachineOperation
```
bcho, line 43: What will be the right ordering between repave and machine operation? I.e. for node/machine X, I queued a soft restart machine operation, hard restart machine operation, then update the machine to use new MCV and delete the node object. Or do we just block when there are pending operations / missing node (waiting repave)?

docs/content/concepts/new_crds.md
phealy: > the spec fields of the object will become immutable

I planned that the immutability would only apply to the spec fields so the controllers could still update the status.

docs/content/concepts/new_crds.md
phealy: I'm not concerned with indicating the applied settings version - we already have an annotation for that (just the regular MachineConfigurationVersion). My concern is the following:

If we indicate we want a repave to take effect by deleting the node object, but we have some types of changes that can be applied live without rebuilding the whole node (maybe a containerd setting or something), how can the user indicate they want that setting to be applied to the node?

I think an applyConfiguration operation would work here for this - it could support repave as well by just having the controller do the cordon/drain/delete node if it's a repave-required operation.

docs/content/concepts/new_crds.md
phealy: I think just returning a 409 conflict would work here instead of having ordering take effect - a webhook on operations that checks "are there any non-terminal-state operations object for this node? if so, reject with 409." would be simple and easy to understand.

docs/content/concepts/new_crds.md
bcho: @phealy  so if no active machines running under the MCV, then its deployed status should be false?

jveski: If we don't care about the `deletionTimestamp` state a finalizer might work. But otherwise a VAP seems reasonable.

docs/content/concepts/new_crds.md
jveski: If we expect it to be used frequently it makes sense as a dedicated spec field. But if it's a kind of necessary evil that we hope folks won't use then using annotations might imply that intention. IMO it's fine either way though

docs/content/concepts/new_crds.md
jveski: I think it's still nice to have. Otherwise the client would do something like this:

- wait for the operation to reach a terminal state
- check the status of the Machine to see if a new enough config was applied

If the second read is from an informer cache, it's always possible that the informer is stale and the client gets a machine status from _before_ the operation completed. Unless I'm missing something about the machine/node/operation relationship.

docs/content/concepts/new_crds.md
phealy: Two fields:

deployedNodes: <int> number of machines currently using this profile
deployed: false when originally created, changes to true once it has been deployed once, never changes back. This is basically the "locked" field, where once a config has been deployed it can no longer be changed.

phealy: I'd prefer a VAP over having something stuck in deletion permanently.

bcho (approved):
per discusison, we will merge this design doc now as we have reached agreement on the direction. Further refinements will be added during implementation.

PR #60, opened 2026-04-21
Cachildr/more inventory

cchildress:
* Always fetch disk serial number from udevadm.
* Rename `inventory` to `inventory-agent`.
* Add `inventory-aggregator` to accept centralized data from agents.
* Add `inventory-viewer` to browse the inventory DB.
* Add `inventory-inspector` stub to be ready to look for conflicting inventory data.
* Add Kubernetes manifests and build info for centralized components.


cmd/inventory/inventory-collector/main.go
```diff
@@ -0,0 +1,69 @@
...
+
+	inventorycollector "github.com/Azure/unbounded-kube/internal/inventory/collector"
+)
+
+func main() {
+	config := inventorycollector.Config{}
+
+	rootCmd := &cobra.Command{
+		Use:   "inventory-collector",
+		Short: "Collect and store node inventory data",
```
jwilder: inventory-agent and inventory-collector sound like they do the same thing at first glance.  Could we rename inventory-collector to inventory-service or something that implies a server side component?

cmd/inventory/inventory-agent/main.go
```diff
@@ -0,0 +1,34 @@
...
+	rootCmd := &cobra.Command{
+		Use:   "inventory",
+		Short: "Collect node inventory data",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return inventory.Execute(config)
+		},
+	}
+
+	rootCmd.Flags().BoolVar(&config.Debug, "debug", false, "Enable debug output")
+	rootCmd.Flags().StringVar(&config.DbPath, "db", "./inventory.db", "Path to the output database file")
```
jwilder, line 27: I'm assuming the agent stores locally and then periodically sends this state to the server?

cmd/inventory/inventory-agent/main.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	inventory "github.com/Azure/unbounded-kube/internal/inventory/agent"
+)
+
+func main() {
```
jwilder, line 15: I wonder if the agent should eventually be merged into the unbounded-agent so there is only one binary that is necessary to run.  With a second binary, we have to think about how it gets updated for patching/feature changes.

cmd/inventory/inventory-inspector/main.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Azure/unbounded-kube/internal/inventory/inspector"
+)
+
+func main() {
```
jwilder, line 15: Can you add some docs to explain what each binary is?  How is this one different from the viewer?

jwilder (dismissed):

cmd/inventory/inventory-agent/main.go
cchildress: It does store locally (that `inventory.db` file by default). It does a write once operation to the remote server. The way to collect data periodically would be to schedule the agent to run on a cron or systemd timer.

cmd/inventory/inventory-agent/main.go
cchildress: Yeah, I don't think that would be a problem. We'd decided to approach them separately for right now to just avoid stepping on toes. I'm going to say that's a future PR whenever we're ready to tackle that.

cmd/inventory/inventory-collector/main.go
cchildress: Sure, it's `inventory-aggregator` now.

cmd/inventory/inventory-inspector/main.go
cchildress: Yup, good idea.

api/inventory/v1/inventory.proto
```diff
@@ -0,0 +1,42 @@
...
+  // JSON-encoded device attributes.
+  bytes attributes = 5;
+}
+
+// NeighborRecord represents a discovered network neighbor.
+message NeighborRecord {
+  string host_identifier = 1;
+  string local_interface = 2;
+  // JSON-encoded neighbor attributes.
+  bytes attributes = 3;
```
plombardi89, line 22: Is there no schema to this? Attributes is just a blob of whatever?

deploy/inventory/collector/01-deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+                || psql "host=$POSTGRES_HOST port=$POSTGRES_PORT user=$POSTGRES_USER password=$POSTGRES_PASSWORD dbname=postgres sslmode=$POSTGRES_SSL_MODE" \
+                  -c "CREATE DATABASE \"$POSTGRES_DB_NAME\""
+          envFrom:
+            - configMapRef:
+                name: inventory-database-config
+            - secretRef:
+                name: pg-creds
+      containers:
+        - name: inventory-collector
+          image: stargatetmedev.azurecr.io/inventory-collector:latest
```
plombardi89, line 36: These are going to need to get stamped with the GHCR repo and version just like the other manifests

deploy/inventory/collector/01-deployment.yaml
```diff
@@ -0,0 +1,41 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: inventory-collector
+  namespace: inventory
```
plombardi89, line 6: Namespace will need to configurable. Take a look at the net and controller manifests for how we're rendering the YAML from templates at build time

plombardi89: We should probably just put this in the unbounded namespace by default as well rather than introducing yet another namespace.

images/inventory/aggregator/Containerfile
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# Build stage
+FROM mcr.microsoft.com/azurelinux/base/core:3.0@sha256:35149ae8dd179684f969944f54a337c665a64e702486154eb44253fb39c2505b AS builder
```
plombardi89: If you can get these on ubuntu now you'd be saving me a headache later.

api/inventory/v1/inventory.proto
cchildress: Yup, that's expected. The top level schema is generic to any and all devices. We store device specific attributes in a json blob.

deploy/inventory/collector/01-deployment.yaml
cchildress: Yup, all set.

images/inventory/aggregator/Containerfile
cchildress: Good idea.

deploy/inventory/collector/01-deployment.yaml
cchildress: I'll grab this one in a new PR.

jwilder (approved):

PR #76, opened 2026-04-24
Extract agent bootstrap library into pkg/agent/ packages

bcho:
## Summary

Extract the agent's bootstrap logic from `cmd/agent/internal/` into importable packages under `pkg/agent/` so that external consumers (e.g. AKS Flex Node) can use them as a shared Go library.

- Move goalstates, phases (host/rootfs/nodestart/nodestop/reset), utilio, utilexec to `pkg/agent/`
- Introduce `pkg/agent/config/` with IR config types (`AgentConfig`, `AgentClusterConfig`, `AgentKubeletConfig`)
- Move TPM attestation to `cmd/agent/internal/attest/` (not part of public API)
- Add `UnboundedAgentConfig` in `internal/provision` embedding `config.AgentConfig` + `Attest` for backward compatibility
- Type aliases in `internal/provision` ensure zero churn for existing controller code

## Exec credential plugin support

- Add `KubeletAuthInfo` type with `BootstrapToken` and `ExecCredential` (mutually exclusive)
- `ExecCredential` uses upstream `k8s.io/client-go/tools/clientcmd/api.ExecConfig` type directly
- Replace Go text template kubeconfig rendering with type-safe `clientcmd.Write()`
- When exec credential is configured, kubeconfig is written directly to kubelet kubeconfig path (no TLS bootstrap); `--bootstrap-kubeconfig` flag is omitted from the systemd drop-in
- Certificate rotation (`--rotate-certificates`) is preserved for both auth modes
- Delete `bootstrap-kubeconfig.yaml` template (replaced by `clientcmd.Write`)

## CRI/CNI version overrides

- Add `CRIConfig` (containing `ContainerdConfig` and `RuncConfig`) and `CNIConfig` to `AgentConfig`
- `ResolveMachine` prefers config-provided versions over hardcoded constants, falling back to library defaults when unset
- Enables consumers like AKS Flex Node to pass through their own containerd, runc, and CNI plugin versions

## Public API surface

```
pkg/agent/config      - AgentConfig IR types (versioning TBD, consumers define own public config)
pkg/agent/goalstates  - ResolveMachine, RootFS, NodeStart, Containerd, Kubelet
pkg/agent/phases      - Task interface, Serial, Parallel, ExecuteTask
pkg/agent/phases/host - InstallPackages, ConfigureOS, ConfigureNFTables, DisableDocker, DisableSwap
pkg/agent/phases/rootfs - DownloadKubeBinaries, DownloadCRIBinaries, DownloadCNIBinaries, Provision
pkg/agent/phases/nodestart - StartNode, ConfigureKubelet, ConfigureContainerd
pkg/agent/phases/nodestop - StopNode
pkg/agent/phases/reset - CleanupMachine
pkg/agent/utilio      - Download and file utilities
pkg/agent/utilexec    - Command execution utilities
```

## What stays internal

- `cmd/agent/internal/cmd/` - CLI commands
- `cmd/agent/internal/daemon/` - drift detection daemon
- `cmd/agent/internal/attest/` - TPM attestation (unbounded-specific)
- `internal/provision` - AgentAttestConfig, UnboundedAgentConfig, BuildAgentConfig


plombardi89:
Can we put it under `pkg/agent` so the project root doesn't get wild and crazy?

plombardi89:
I don't think we should export stuff like the various util packages. We should be pretty choosy about what functions and types we export for public consumption.

bcho:
> I don't think we should export stuff like the various util packages. We should be pretty choosy about what functions and types we export for public consumption.

in aks flex node we do have some scenarios where we need to reuse the utilio part, but agree with you we should minimize the exported functions/interfaces. Will do a clean up round after getting the aks flex flow running e2e

bcho:
> I don't think we should export stuff like the various util packages. We should be pretty choosy about what functions and types we export for public consumption.

moved them to internal package for now

pkg/agent/goalstates/kubelet.go
```diff
@@ -3,15 +3,17 @@
...
 	// KubeletBinPath is the absolute path to the kubelet binary inside the
 	// machine container (e.g. /usr/local/bin/kubelet).
 	KubeletBinPath string
 
-	// BootstrapToken is the bootstrap token in "<token-id>.<token-secret>"
-	// format used for TLS bootstrapping with the API server.
-	BootstrapToken string
+	// KubeletAuthInfo holds the authentication configuration for the
+	// kubelet. Exactly one of BootstrapToken or ExecCredential must be set.
+	config.KubeletAuthInfo
```
imiller31, line 16: Why are switching to an embedded struct here?

imiller31 (approved):

pkg/agent/goalstates/kubelet.go
bcho: mostly we don't need to update all existing references from `x.BootstrapToken` to `x.KubeletAuthInfo.BootstrapToken`

PR #59, opened 2026-04-20
Add Operation CRD and agent daemon operation controller with soft-reboot support

bcho:
## Summary

Replace the temporary ConfigMap-based operation shim with a proper Operation custom resource (`unbounded-kube.io/v1alpha3`) for managing machine operations like soft-reboot.

## Changes

- **Operation CRD**: New cluster-scoped CR with `SoftReboot` and `HardReboot` types, status subresource, owner references for GC, and TTL-based cleanup
- **Agent daemon**: Rewrite operation watcher to watch Operation CRs instead of ConfigMaps; replace `reconciler_softrestart` with generic `reconciler_operation` that dispatches by type
- **kubectl plugin**: Rewrite `soft-reboot` command to create an Operation CR with owner reference and TTL (default 300s), then watch status until completion
- **RBAC**: Replace ConfigMap rules with `operations` and `operations/status` verbs
- **Tests**: 7 operation reconciler tests covering all phases, TTL cleanup, and edge cases

## E2E Validation

Validated on `oc-vm3` (Ubuntu 24.04, Standard_D2as_v7):

```
$ kubectl get machines -o wide
NAME    HOST   PHASE   K8S VERSION   AGE
agent                                14m

$ ./bin/kubectl-unbounded machine soft-reboot agent
  --> Soft-rebooting Machine agent...
  operation         agent-softreboot-1776734492
  --> Operation SoftReboot: agent-softreboot-1776734492 in progress...
  --> Operation SoftReboot: agent-softreboot-1776734492 completed
  ready

$ kubectl get operations -o wide
NAME                          MACHINE   TYPE         PHASE       AGE
agent-softreboot-1776734492   agent     SoftReboot   Completed   4m5s

$ kubectl get operations -o yaml
apiVersion: unbounded-kube.io/v1alpha3
kind: Operation
metadata:
  name: agent-softreboot-1776734492
  ownerReferences:
  - apiVersion: unbounded-kube.io/v1alpha3
    kind: Machine
    name: agent
    uid: d5fc240c-0b2a-4a3f-9468-1d2aabf691bb
spec:
  machineRef: agent
  ttlSecondsAfterFinished: 300
  type: SoftReboot
status:
  completedAt: "2026-04-21T01:21:33Z"
  phase: Completed
  startedAt: "2026-04-21T01:21:32Z"
```

Node returned to `Ready` after soft reboot completed successfully.

PR #96, opened 2026-04-28
Add MachineOperation, MachineConfiguration, and MachineConfigurationVersion CRDs

bcho:
## Summary

Adds three new custom resource definitions following the design in [New CRDs](https://github.com/Azure/unbounded/blob/main/docs/content/concepts/new_crds.md), derived from the prototype in #59:

- **MachineOperation** (`mop`): discrete operations (SoftReboot, HardReboot) targeting machines
- **MachineConfiguration** (`mc`): deployment-like configuration profiles with update strategies and machine selectors
- **MachineConfigurationVersion** (`mcv`): immutable versioned snapshots of configuration templates
- Extends the **Machine** CR with `configurationRef` (spec), `configuration` status, `NodeUpdated` and `ConfigurationPending` conditions, and `AnnotationConfigurationVersion`
- Includes generated deepcopy methods and CRD manifests

This PR contains only the CR type definitions and generated artifacts. Controller, CLI, and agent implementations will follow in subsequent PRs.


api/machina/v1alpha3/machineoperation_types.go
```diff
@@ -0,0 +1,151 @@
...
+	OperationPhaseComplete   OperationPhase = "Complete"
+	OperationPhaseFailed     OperationPhase = "Failed"
+)
+
+// MachineOperationSpec defines the desired state of a MachineOperation.
+type MachineOperationSpec struct {
+	// MachineRef is the name of the Machine CR this operation targets.
+	// Either machineRef or machineSelector must be set.
+	// +optional
+	MachineRef string `json:"machineRef,omitempty"`
```
imiller31, line 86: Thinking about the VM operations here, so we have an assumption that Machine CR will contain necessary identifying information for the operations?

e.g. ARM ResourceID and Tenant for Azure, Proxmox VMID and pve endpoint, etc.

api/machina/v1alpha3/machineoperation_types.go
```diff
@@ -0,0 +1,151 @@
...
+	// +optional
+	Parameters map[string]string `json:"parameters,omitempty"`
+
+	// TTLSecondsAfterFinished limits the lifetime of a completed or failed
+	// MachineOperation. If set, the agent deletes the MachineOperation
+	// this many seconds after it reaches a terminal phase. If unset, the
+	// MachineOperation is kept indefinitely.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
```
imiller31, line 114: I wonder if each operation should have TTL as well. It feels like we could get an operation stuck in a Pending state pretty easily

api/machina/v1alpha3/machineoperation_types.go
```diff
@@ -0,0 +1,151 @@
+// Copyright (c) Microsoft Corporation.
```
imiller31, line 1: Do we need some type of locking mechanism on these operations for controllers reconciling based off this CR?

Probably not needed for this PR, but what happens if we have multiple controllers and controller replicas?

api/machina/v1alpha3/machineoperation_types.go
bcho: yeah i think we will need a way to correlate the cloud provider identity in machine side. Or we can correlate with the cloud provider id from node then to machine

api/machina/v1alpha3/machineoperation_types.go
bcho: yep i think we can only allow one ongoing operation per machine based on the previous conv https://github.com/Azure/unbounded/pull/46#discussion_r3119610303 so in this case we just need to leverage some sort of leader election to run one operator inside the cluster at a time

api/machina/v1alpha3/machineoperation_types.go
imiller31: We could probably label the machine with this info or put it into the CR. I'm not sure what is the most ergonomic, but we do need some way of mapping a given machine to it's underlying host.

I guess how does the inventory piece fit in here?

@jveski @cchildress might have some more info

imiller31 (approved):

api/machina/v1alpha3/machineoperation_types.go
bcho: discussed offline, we will implement some sort of imds + per site provider approach to provide the infos

api/machina/v1alpha3/machineoperation_types.go
bcho: yeah i think we can add one in the future if this is needed, let's start without it for now

PR #98, opened 2026-04-28
Add machine operations controller

imiller31:
## Summary
The goal of this PR is to introduce an external operations controller that can handle Machine Operations that must be invoked externally from the machine. Hard Reboot, PowerOff (not shutdown), and PowerOn are the initially supported operations.

This PR focuses on Azure VMs assuming that the identity is available via the DefaulCredential flow. We likely will need to expand this more to support cross-tenant authentication.

- add MachineOperation, MachineConfiguration, and MachineConfigurationVersion CRDs, including Azure external VM identity on Machine
- add standalone machine-ops-controller with Azure VM HardReboot, PowerOff, and PowerOn support
- add machine-ops manifests, image build targets, docs, and Azure E2E harness

Note: This PR builds off of #96 
## Validation
- manual Azure E2E: HardReboot, PowerOff, PowerOn, and TTL cleanup against two joined Azure VMs


api/machina/v1alpha3/machine_types.go
```diff
@@ -123,6 +129,11 @@ type MachineSpec struct {
 	// +optional
 	Operations *OperationsSpec `json:"operations,omitempty"`
 
+	// External contains configuration for controllers that operate on the
+	// machine from outside the host OS, such as cloud VM APIs or BMCs.
+	// +optional
+	External *ExternalMachineSpec `json:"external,omitempty"`
```
imiller31: I'm not super tied to this pattern, but I'm unsure what the most ergonomic way to do this is. For each externally managed machine (VM), we need enough identifying information to target a specific cloud provider with correct auth.

deploy/machine-ops/03-deployment.yaml.tmpl
```diff
@@ -0,0 +1,112 @@
...
+        - name: controller
+          image: {{ default "ghcr.io/azure/machine-ops-controller:latest" .ControllerImage }}
+          imagePullPolicy: Always
+          args:
+            - --leader-elect=true
+            - --leader-elect-namespace={{ default "unbounded-kube" .Namespace }}
+            - --max-concurrent-reconciles={{ default "10" .MaxConcurrentReconciles }}
+            - --metrics-bind-address=:8080
+            - --health-probe-bind-address=:8081
+            {{- if .OCIConfigSecretName }}
```
imiller31, line 61: This is pretty much a hack right now until I figure out a more elegant solution for loading many different authorization mechanisms

deploy/machina/crd/unbounded-cloud.io_machines.yaml
```diff
@@ -241,6 +241,47 @@ spec:
                 required:
                 - name
                 type: object
+              external:
+                description: |-
+                  External contains configuration for controllers that operate on the
+                  machine from outside the host OS, such as cloud VM APIs or BMCs.
+                properties:
+                  azureVM:
```
bcho: if we are just using resource id, can we simply follow the node provider id setup?

```
spec:
  providerID: azure:///subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.Compute/virtualMachineScaleSets/<vmss-name>/virtualMachines/0
```

basically just `<provider-scheme>://<provider-local-id>`. In this case we can simply copy the provider id value from node object. The node provider id value is populated by cloud provider.

imiller31: @bcho I think we will need some way to encode tenant ID and reference the auth mechanism in the future (should we use MSI or some secret in the cluster for auth?). 

I guess we could map the specific provider identifiers to a clusterwide config. I don't have a strong opinion here, so I think we just need a consensus on what would be the most ergonomic.

bcho: can those per cloud / site setups stay in machine configuration or site settings? Similar to node class in karpenter: https://karpenter.sh/docs/concepts/nodeclasses/

imiller31: I switched the provider string with external moved to the machineconfiguration crds for now. We can iterate on this if we feel like this isn't the right place to store the  info

bcho (approved):

api/machina/v1alpha3/machineconfiguration_types.go
```diff
@@ -91,6 +91,27 @@ type MachineConfigurationTemplate struct {
...
+)
+
+// MachineConfigurationExternal selects the external provider used for
+// out-of-band machine operations.
+type MachineConfigurationExternal struct {
+	// Provider identifies the external control provider for machines using this
+	// configuration.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=AzureVM;OCIInstance
+	Provider string `json:"provider"`
```
jwilder: In prior discussions, I thought we were just going to put Provider on the Machine directly and the controllers would manage machines that matched their Provider.  @plombardi89 @jveski

api/machina/v1alpha3/machineoperation_types.go
```diff
@@ -66,6 +58,12 @@ const (
 	// via BMC (e.g. Redfish). Handled by the machina controller or cloud
 	// controller.
 	OperationHardReboot OperationKind = "HardReboot"
+
+	// OperationPowerOff powers off the host through an out-of-band provider.
+	OperationPowerOff OperationKind = "PowerOff"
```
jwilder: I wonder if we should add a prefix for Host vs Machine operations.  For example, `HostPowerOff` would power off the physical host but `MachinePowerOff` would power off the nspawn machine.  `SoftReboot` becomes `MachineReboot` and `HardReboot` becomes `HostReboot`.  Might make it easier to determine what scope the operation applies.

cmd/machine-ops-controller/main.go
```diff
@@ -0,0 +1,124 @@
...
+
+			return run(ctx, cfg)
+		},
+		Version: version.Version + " (commit: " + version.GitCommit + ")",
+	}
+
+	cmd.Flags().StringVar(&cfg.metricsAddr, "metrics-bind-address", ":8080", "Address for metrics endpoint")
+	cmd.Flags().StringVar(&cfg.probeAddr, "health-probe-bind-address", ":8081", "Address for health probes")
+	cmd.Flags().BoolVar(&cfg.leaderElection, "leader-elect", true, "Enable leader election")
+	cmd.Flags().StringVar(&cfg.leaderElectionNamespace, "leader-elect-namespace", "unbounded-kube", "Namespace for the leader election lease")
```
jwilder, line 47: Did we change the namespace to `unbounded` when we renamed the repo?  @copilot check what is used for the other namespaces definitions.

cmd/machine-ops-controller/main.go
```diff
@@ -0,0 +1,124 @@
...
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+
+	unboundedv1alpha3 "github.com/Azure/unbounded/api/machina/v1alpha3"
+	"github.com/Azure/unbounded/internal/machineops"
+	"github.com/Azure/unbounded/internal/machineops/providers/azurevm"
+	"github.com/Azure/unbounded/internal/machineops/providers/ociinstance"
```
jwilder, line 25: In prior discussions, we had said that a user would run a controller per provider so that it was easy to extend to new providers without having to modify one uber controller.  Is this more of a reference implementation or a convenient way to provide support for some known providers easily?  Having this wouldn't preculde someone from implementing their own "foo-machine-ops-controller", right?

imiller31: [In the reference design](https://github.com/Azure/unbounded/blob/main/docs/content/concepts/new_crds.md), there's a single controller. We did discuss that we likely need a controller per cloud + tenant combination to do in-cloud authentication (if available). For example a TME controller would need to run on a TME machine to use MSI while a Corp tenant controller would need to run on a corp machine. 

The goal here is to have something working with our known providers. The controller can be configured to only operate on specific machines/sites/providers once we have a consensus on where that data should live. 

> Having this wouldn't preclude someone from implementing their own "foo-machine-ops-controller"

Correct, really we just need to determine a way for these controllers to filter out machines that they shouldn't be operating on

api/machina/v1alpha3/machineconfiguration_types.go
imiller31: Yea I'm unclear where this data should live. We ultimately will need info for:
1. Provider
2. Provider Identification (resource ID)
3. Tenant 

Each controller will need to use a combination of Provider and Tenant to determine how to auth to the corresponding cloud.

jwilder (approved):

api/machina/v1alpha3/machineconfiguration_types.go
jwilder: I lean towards just extending the Machine CRD.

api/machina/v1alpha3/machineoperation_types.go
plombardi89: Using the `Machine` prefix here is really confusing because a Machine is almost always thought of as a lower-level primitive in the infrastructure. The fact `systemd-nspawn` refers to them as such as kind of leaking an implementation detail up to userland.

My recommendation would be to use `NodeReboot` as it maps cleanly to the `v1.Node` concept in Kubernetes. We operate on Hosts (the virtual or physical infrastructure) or Nodes (the Kubernetes abstraction and its wrapper, in this case `systemd-nspawn`).

cmd/machine-ops-controller/main.go
plombardi89: We have `unbounded-kube` and `unbounded-net`. Do we want to unify everything into a single namespace or just renamed `unbounded-kube` to `unbounded`?

cmd/machine-ops-controller/main.go
plombardi89: Label on the Machine CR? The controllers can be booted up either knowing of a standard label they handle or given a set of custom labels to watch for that way you can run different controllers even in the same site for different sets of nodes if you need to do it for some reason?

api/machina/v1alpha3/machineconfiguration_types.go
imiller31: Done

cmd/machine-ops-controller/main.go
imiller31: I think we can easily change this later

api/machina/v1alpha3/machineoperation_types.go
imiller31: done

PR #105, opened 2026-04-29
Implement unbounded CLI experience for PR #96

bcho:
## Summary
- Implements the unbounded CLI experience for the MachineOperation and MachineConfiguration APIs introduced in PR #96.
- Adds MachineConfiguration create/get/versions/assign commands under `machine config`.
- Adds MachineOperation-backed machine soft-reboot and hard-reboot commands.

## CLI Experience
- `kubectl unbounded machine config create NAME` creates a MachineConfiguration with optional Kubernetes version, agent image, node labels, taints, update strategy, and priority.
- `kubectl unbounded machine config get [NAME]` lists MachineConfigurations or shows details for one configuration.
- `kubectl unbounded machine config versions NAME` lists MachineConfigurationVersions for a configuration.
- `kubectl unbounded machine config assign MACHINE_NAME --config=CONFIG_NAME [--version=N]` assigns a configuration to a Machine, optionally pinned to a specific version.
- `kubectl unbounded machine soft-reboot NAME [--ttl=SECONDS]` creates and watches a SoftReboot MachineOperation for agent-managed in-place reboot.
- `kubectl unbounded machine hard-reboot NAME [--ttl=SECONDS]` creates and watches a HardReboot MachineOperation for controller-managed host power-cycle reboot.


cmd/kubectl-unbounded/app/cmd_helpers.go
```diff
@@ -24,3 +28,39 @@ func getKubeconfigPath(p string) string {
...
+		t := parseTaint(s)
+		taints = append(taints, t)
+	}
+	return taints
+}
+
+// parseTaint parses a single taint string. Accepted formats:
+//   - key=value:Effect
+//   - key:Effect
+func parseTaint(s string) corev1.Taint {
```
plombardi89: nit: we could take this code from upstream which also has validation for the taint

https://github.com/kubernetes/kubernetes/blob/master/pkg/util/taints/taints.go#L38

cmd/kubectl-unbounded/app/config_assign.go
```diff
@@ -0,0 +1,100 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	v1alpha3 "github.com/Azure/unbounded/api/machina/v1alpha3"
+)
+
+func configAssignCommand() *cobra.Command {
+	var version int32
+
+	cmd := &cobra.Command{
+		Use:   "assign CONFIG_NAME MACHINE_NAME",
```
plombardi89: Not a huge fan of positional args for parameters like this. It's so easy to forget the order. Would be better if we stuck to letting the machine name be the arg but use flag parameters for the other stuff.

cmd/kubectl-unbounded/app/machine_reboot.go
```diff
@@ -39,6 +39,10 @@ func runReboot(ctx context.Context, c client.WithWatch, name string) error {
 		return err
 	}
 
+	if machine.Spec.PXE == nil || machine.Spec.PXE.Redfish == nil {
```
plombardi89, line 42: Are we sure this correct? I would expect a VM in say OCI to be able to hard reboot just via API

plombardi89 (dismissed):

cmd/kubectl-unbounded/app/config_assign.go
bcho: something like `machine config assign <machine> --config <config>` ?

cmd/kubectl-unbounded/app/machine_reboot.go
bcho: I think after we implementing the `hard reboot` in https://github.com/Azure/unbounded/blob/f48f5d136e468e4d00f54e63e69da0dbd8ef59b3/cmd/kubectl-unbounded/app/machine_hard_reboot.go , this reboot implementation can be removed. Then we just need to implement site wise controller for pxe / redfish vm. @jveski / @imiller31

cmd/kubectl-unbounded/app/cmd_helpers.go
bcho: copied the upstream impl to here

cmd/kubectl-unbounded/app/machine_reboot.go
imiller31: I think metalman should ultimately handle all of the pxe / redfish ops

bcho: https://github.com/Azure/unbounded/issues/112 recorded an issue for this, will work on this later

imiller31 (dismissed):

imiller31 (approved):

PR #107, opened 2026-04-30
ebpf: pin cross-site supernet RoutedCidrs to a single nexthop

vpatelsj:
Gateway peers in a multi-gateway pool all advertise the same cross-site supernet (e.g. cluster pod CIDR 10.244.0.0/16). The BPF tunnel-endpoint LPM map programmed each peer as an additional nexthop for that supernet, making the supernet entry ECMP-eligible. The BPF dataplane's flow_hash is direction-asymmetric, so a SYN and its SYN-ACK can hash onto different gateway nexthops; when the SYN-ACK egresses on a gateway that did not see the SYN, that gateway has no flow state and the reply is black-holed. This appeared as a ~25-30% TCP success rate from cluster pods to remote pods that arrived via the supernet fallback route.

Per-peer PodCIDRs are unique to one gateway and remain multi-nexthop ECMP via addPeerBPFEntries; only the supernet RoutedCidrs are pinned to a single deterministic nexthop. Gateway peers are sorted by Name before iteration so the winning nexthop is reproducible across reconciles.

Adds unit tests covering: multi-nexthop append still works for repeated CIDRs (PodCIDR semantics), single-nexthop first-wins, and the single-nexthop helper does not collapse an existing multi-nexthop entry for the same CIDR.

Verified on a live cluster: spark-2c24's BPF map for 10.244.0.0/16 now shows exactly one nexthop (was two), per-peer /24 PodCIDR entries still have multiple nexthops where expected.

### DIFF
--- a/cmd/unbounded-net-node/ebpf_geneve_config.go
+++ b/cmd/unbounded-net-node/ebpf_geneve_config.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"k8s.io/klog/v2"
@@ -329,7 +330,10 @@ func configureEBPFTunnelPeers(
 		addPeerBPFEntries(bpfEntries, peer.PodCIDRs, underlayIP, uint32(cfg.GeneveVNI), peerIfIdx, peerFlags, peerProto, peer.Name)
 	}
 
-	for _, gwPeer := range gatewayPeers {
+	sortedGatewayPeers := append([]gatewayPeerInfo(nil), gatewayPeers...)
+	sort.Slice(sortedGatewayPeers, func(i, j int) bool { return sortedGatewayPeers[i].Name < sortedGatewayPeers[j].Name })
+
+	for _, gwPeer := range sortedGatewayPeers {
 		if len(gwPeer.InternalIPs) == 0 || len(gwPeer.PodCIDRs) == 0 {
 			continue
 		}
@@ -341,7 +345,9 @@ func configureEBPFTunnelPeers(
 
 		peerIfIdx, peerFlags, peerProto := resolveEBPFPeerTarget(gwPeer.TunnelProtocol, geneveIfIndex, ipipIfIndex, defaultIfIdx, cfg)
 		addPeerBPFEntries(bpfEntries, gwPeer.PodCIDRs, underlayIP, uint32(cfg.GeneveVNI), peerIfIdx, peerFlags, peerProto, gwPeer.Name)
-		addPeerBPFEntries(bpfEntries, gwPeer.RoutedCidrs, underlayIP, uint32(cfg.GeneveVNI), peerIfIdx, peerFlags, peerProto, gwPeer.Name)
+		// Supernet RoutedCidrs are pinned to a single nexthop (first peer wins)
+		// to avoid asymmetric ECMP across non-equivalent gateways.
+		addPeerBPFEntriesSingleNexthop(bpfEntries, gwPeer.RoutedCidrs, underlayIP, uint32(cfg.GeneveVNI), peerIfIdx, peerFlags, peerProto, gwPeer.Name)
 	}
 
 	// Store entries for deferred reconcile (after VXLAN and WG entries are added).
@@ -499,7 +505,10 @@ func configureEBPFVXLANPeers(
 		addPeerBPFEntries(bpfEntries, peer.PodCIDRs, underlayIP, vni, tunnelIfIndex, ebpfpkg.TunnelFlagSetKey, ebpfpkg.TunnelProtoVXLAN, peer.Name)
 	}
 
-	for _, gwPeer := range gatewayPeers {
+	sortedGatewayPeers := append([]gatewayPeerInfo(nil), gatewayPeers...)
+	sort.Slice(sortedGatewayPeers, func(i, j int) bool { return sortedGatewayPeers[i].Name < sortedGatewayPeers[j].Name })
+
+	for _, gwPeer := range sortedGatewayPeers {
 		if len(gwPeer.InternalIPs) == 0 || len(gwPeer.PodCIDRs) == 0 {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


vpatelsj:
## Problem

On `vapa-unbounded-2` (AKS + remote DGX Spark / lab gateway), TCP from cluster pods to remote pods succeeded only ~25-30% of the time. The same destination, called repeatedly from a single client pod, would alternate between immediate success and connect timeouts. Re-running the probe always reproduced.

## Observation

Captured SYN-by-SYN on all three hops (AKS node, lab gateway, remote spark) using hostNetwork tcpdump pods with `kubectl logs` (since `kubectl exec` was unreliable on the spark). The smoking gun was on the spark, comparing two adjacent probes:

**Failing flow (probe 6, src port 47580):**

```
wg51822 In  SYN     10.244.1.11.47580 > 10.245.2.40.11434       # SYN arrives via wg51822
veth     P  SYN-ACK 10.245.2.40.11434 > 10.244.1.11.47580       # ollama replies
wg51821 Out SYN-ACK 10.245.2.40.11434 > 10.244.1.11.47580       # !!! leaves via wg51821
```

AKS never receives the SYN-ACK. AKS retransmits 4×, every retransmit comes in `wg51822` and every SYN-ACK leaves `wg51821`. AKS times out.

**Succeeding flow (probe 5, src port 47574, ~1s earlier):**

```
wg51822 In  SYN     10.244.1.11.47574 > 10.245.2.40.11434
wg51822 Out SYN-ACK 10.245.2.40.11434 > 10.244.1.11.47574       # leaves via SAME interface
```

AKS receives the SYN-ACK on `geneve0 In` ~40 ms later. Handshake completes.

The gateway-node capture was **empty** — traffic from AKS to the spark pod was not transiting `apollo-lab-bou-gw` at all. So this is a direct AKS-overlay ↔ spark-WG path, and the asymmetry is entirely between the two `wg5182x` interfaces on the spark.

`unroute` on the spark confirmed the BPF tunnel-endpoint LPM map had two equal-priority nexthops for the cluster pod supernet:

```
# spark-3d37 BPF map (BEFORE fix)
10.244.0.0/16  ->  10.224.0.6  wg51821     # vmss000000
10.244.0.0/16  ->  10.224.0.7  wg51822     # vmss000001        <-- duplicate
10.244.0.0/24  ->  10.224.0.7  wg51822     # per-peer PodCIDR (correct)
10.244.3.0/24  ->  10.224.0.6  wg51821     # per-peer PodCIDR (correct)
```

## Root cause

The BPF dataplane's `flow_hash` is direction-asymmetric: a SYN's hash key derives from `(src=client, dst=server, sport, dport, proto)`, while the SYN-ACK's hash key derives from `(src=server, dst=client, sport, dport, proto)`. These two hashes are independent — for an ECMP-eligible LPM entry with N nexthops, each direction independently picks an index `hash mod N`, so a flow can land on different nexthops in each direction. The ~25-30% success rate is consistent with N=2 nexthops where the two hashes happen to agree on a single nexthop choice for ~25% of source ports.

Three callsites in `cmd/unbounded-net-node/ebpf_geneve_config.go` (GENEVE, VXLAN, WireGuard configurators) called `addPeerBPFEntries` for `gwPeer.RoutedCidrs`. That helper appends a nexthop for an existing CIDR key — correct behavior for per-peer PodCIDRs (which are unique per gateway peer), but wrong for cross-site supernets like `10.244.0.0/16` (which every gateway peer in the pool advertises identically). Each gateway peer's `RoutedCidrs` thus contributed an additional nexthop to the supernet LPM entry, producing the multi-nexthop ECMP race.

## Fix

- New helper `addPeerBPFEntriesSingleNexthop` in `ebpf_geneve_config.go`: same shape as `addPeerBPFEntries` but **skips a CIDR if it already has any nexthop programmed** (first-peer-wins). Sets `TUNNEL_F_HEALTHY` so traffic flows immediately rather than waiting for the healthcheck loop.
- The three call sites for `gwPeer.RoutedCidrs` switched from `addPeerBPFEntries` -> `addPeerBPFEntriesSingleNexthop`.
- Gateway peers are sorted alphabetically by `Name` before iteration so the winning nexthop is deterministic and reproducible across reconciles (the same gateway wins after every restart).
- **Per-peer PodCIDRs are unchanged** — they remain multi-nexthop ECMP via `addPeerBPFEntries`. The /24 PodCIDR entries are LPM-preferred over the /16 supernet, so traffic to known per-peer destinations still uses the correct gateway. The single-nexthop pinning only affects the supernet fallback for unknown /16 destinations.

## Verification

3 new unit tests in `cmd/unbounded-net-node/ebpf_routedcidr_dedupe_test.go`:

- `TestAddPeerBPFEntries_MultiNexthopForSameCIDR` — multi-nexthop append still works for repeated CIDRs (PodCIDR semantics)
- `TestAddPeerBPFEntriesSingleNexthop_FirstWins` — single-nexthop helper retains exactly one nexthop set by the first caller
- `TestAddPeerBPFEntriesSingleNexthop_DoesNotAffectExistingMultiNexthop` — calling the single-nexthop helper for a CIDR that already has multi-nexthop entries leaves them untouched

Live cluster after rollout: spark-2c24's BPF map for `10.244.0.0/16` shows exactly one nexthop (was two); per-peer `/24` entries still show their unique nexthop:

```
# spark-2c24 BPF map (AFTER fix)
10.244.0.0/16  ->  10.224.0.6  wg51821     # ONE nexthop (was two)
10.244.0.0/24  ->  10.224.0.7  wg51822     # /24 unchanged
10.244.3.0/24  ->  10.224.0.6  wg51821     # /24 unchanged
```

End-to-end probe `open-webui -> ollama@spark`: 20/20 success after fix (was 5-6/20 before). HTTP `GET /api/tags` against ollama: 5/5 HTTP 200 in ~88 ms.

## Risk

Low. The change only affects gateway `RoutedCidrs` (the cross-site supernet fallback path). Per-peer routes are unchanged. The first-peer-wins choice is deterministic across reconciles thanks to the alphabetical-by-name sort, so behavior is reproducible.

The trade-off vs. ECMP: legitimate ECMP across equivalent gateways for the supernet fallback is now disabled. This is an acceptable trade because (a) the supernet fallback is only used for destinations that don't have a more-specific PodCIDR entry, which should be rare in steady state, and (b) "ECMP across non-equivalent gateways" is exactly what produced the asymmetric routing bug.

phealy:
This actually causes more problems than it solves when you get to multiple gateway pools. We have rules in place in iptables to allow traffic even if ctstate is wrong - ECMP should work fine.

PR #108, opened 2026-04-30
node: flush stale L2-transit UDP conntrack on masquerade reconcile

vpatelsj:
When a node has a Linux bridge interface on a downstream network and net.bridge.bridge-nf-call-iptables is enabled (required for k8s on the same node), every frame that transits the bridge invokes the netfilter hooks even when the node is not the routing target. This is normal for k8s, but creates a hazard: if a downstream host's frames are bridged out the same VLAN (not routed via this node's default-gateway iface), conntrack records the flow as UNREPLIED with no NAT applied because the MASQUERADE rule is scoped to '-o <default-iface>'.

If the downstream host's default route later changes to point at this node, the same flow now SHOULD be SNAT'd. Linux's NAT engine only computes NAT decisions on the first packet of a flow and reuses thereafter, so subsequent packets keep using the no-NAT decision from the cached entry. Packets leave with the wrong source address, the upstream peer never replies, and the affected flow stays broken until the entry ages out (effectively never for keepalived UDP).

Add FlushStaleTransitUDPConntrack which scans the conntrack table and removes UDP entries that meet all of: protocol UDP, original and reply tuples are exact mirrors (no SNAT/DNAT applied), neither tuple end is a local IP. The L2-transit-with-no-NAT case is precisely this shape. TCP is intentionally left alone (handshake state makes staleness less common, and TCP CONNRESET would be more disruptive than the self-recovery wait).

Wired into syncMasqueradeRules so any reconcile (already periodic) will opportunistically clear stale entries within seconds of a downstream route flip. No tunables, no new Site/Node configuration required.

Reproduced on a live cluster: a remote node's default route was flipped from a non-NAT'ing router to the unbounded-net gateway node, which left the gateway's apollo-side conntrack with a stale un-NAT'd WireGuard handshake entry. WG handshakes silently failed until the entry was manually flushed; with this change the next reconcile flushes it automatically.

Adds 7 unit tests for the filter predicate covering matched stale entries, NAT-translated entries, TCP, local-source/dest, nil flow, and port-mismatched mirror tuples.

### DIFF
--- a/cmd/unbounded-net-node/site_watch_reconcile.go
+++ b/cmd/unbounded-net-node/site_watch_reconcile.go
@@ -2110,6 +2110,22 @@ func syncMasqueradeRules(state *wireGuardState, nonMasqCIDRs []string) {
 	} else {
 		klog.V(4).Infof("Synced masquerade rules for default gateway (v4: %s, v6: %s)", defaultIfaceV4, defaultIfaceV6)
 	}
+
+	// Opportunistically flush stale L2-transit UDP conntrack entries. This
+	// guards against the bridge-nf-call-iptables / route-flap class of bug
+	// where a downstream host's pre-route-change L2 frames left untranslated
+	// conntrack entries on this node, causing post-route-change packets to
+	// skip MASQUERADE. See FlushStaleTransitUDPConntrack for details.
+	if defaultIfaceV4 != "" {
+		localIPs, err := unboundednetnetlink.LocalIPv4Addresses()
+		if err != nil {
+			klog.V(3).Infof("Failed to enumerate local IPs for conntrack flush: %v", err)
+		} else if n, err := unboundednetnetlink.FlushStaleTransitUDPConntrack(localIPs); err != nil {
+			klog.V(3).Infof("Failed to flush stale transit conntrack: %v", err)
+		} else if n > 0 {
+			klog.V(2).Infof("Flushed %d stale L2-transit UDP conntrack entries", n)
+		}
+	}
 }
 
 // logMeshPeerChanges logs which mesh peers were added or removed
--- a/internal/net/netlink/conntrack_flush.go
+++ b/internal/net/netlink/conntrack_flush.go
@@ -1,0 +1,137 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package netlink
+
+import (
+	"fmt"
+	"net"
+	"syscall"
+
+	"github.com/vishvananda/netlink"
+)
+
+// FlushStaleTransitUDPConntrack scans the conntrack table and removes UDP
+// entries that are pure L2 transit and have not been NAT-translated.
+// It returns the number of entries removed.
+//
+// Specifically, an entry is removed when all of the following hold:
+//   - protocol is UDP
+//   - the original and reply tuples are exact mirrors (no SNAT/DNAT applied)
+//   - neither the original source nor the original destination is one of
+//     the supplied local IPs
+//
+// This addresses a class of bugs caused by Linux bridges with
+// bridge-nf-call-iptables=1: when a downstream host's L2 frames transit a
+// bridge but are not routed by the bridge node, conntrack still tracks the
+// flow as UNREPLIED with no NAT applied. If the downstream host's default
+// route later changes to point at the bridge node, the same flow now SHOULD
+// be SNAT'd by an existing MASQUERADE rule, but conntrack reuses the stale
+// (un-NAT'd) entry and skips MASQUERADE -- packets leave with the wrong
+// source address and the upstream peer never replies.
+//
+// Periodically calling this function clears such stale entries so the next
+// packet for that flow takes a fresh path through the netfilter hooks.
+// UDP-only is intentional: TCP flows reach ESTABLISHED via SYN/SYN-ACK and
+// would be safe to leave; the staleness mostly affects connectionless UDP
+// (WireGuard handshake, GENEVE keepalive, etc).
+func FlushStaleTransitUDPConntrack(localIPs []net.IP) (uint, error) {
+	filter := newStaleTransitUDPFilter(localIPs)
+
+	// IPv4 only for now: conntrack is per-family in netlink, and the bridge
+	// transit issue is overwhelmingly an IPv4 NAT problem.
+	n, err := netlink.ConntrackDeleteFilter(netlink.ConntrackTable, syscall.AF_INET, filter)
+	if err != nil {
+		return 0, fmt.Errorf("conntrack delete: %w", err)
+	}
+
+	return n, nil
+}
+
+// staleTransitUDPFilter implements netlink.CustomConntrackFilter and matches
+// stale L2-transit UDP entries (see FlushStaleTransitUDPConntrack).
+type staleTransitUDPFilter struct {
+	localIPs map[string]struct{}
+}
+
+func newStaleTransitUDPFilter(localIPs []net.IP) *staleTransitUDPFilter {
+	set := make(map[string]struct{}, len(localIPs))
+	for _, ip := range localIPs {
+		if ip == nil {
+			continue
+		}
+
+		set[ip.String()] = struct{}{}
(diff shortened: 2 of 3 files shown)


vpatelsj:
## Problem

Even after fixing the eBPF supernet single-nexthop issue (sister PR), end-to-end probe `open-webui -> ollama@spark-3d37` was still 0/20. Investigation revealed a separate bug: a remote node's WireGuard tunnel to one of the AKS gateway VMSSes was completely dead, but the daemon was unaware (the WG config and BPF map looked correct on both sides).

## Observation

On the remote spark, `wg show wg51821` reported the peer was up:

```
peer: QvrMg1PvQXTbiuaI73Kdm0W2x9xjxLhPFo4IIZsDl3s=     # vmss000000
  endpoint: 52.138.2.134:51820
  allowed ips: 10.224.0.0/16, 10.244.0.0/16
  transfer: 0 B received, 5.91 MiB sent                # !!! one-way: only TX
```

`link_stats_monitor` flagged steady tx_dropped on `wg51821`:

```
Link stats: detected 1 interface(s) with incrementing error counters
  interface /wg51821: tx_dropped +70
  interface /wg51821: tx_dropped +49
  interface /wg51821: tx_dropped +66
```

Meanwhile on vmss000000:

```
peer: dh9nqMSqRU0NBrBWJKJMLYfXO9hfrP1NiOHMxBA13ls=     # spark-3d37
  endpoint: 157.58.218.71:47033                         # stale source IP
  allowed ips: 10.245.2.0/24, 172.18.10.7/32
  transfer: 141.93 KiB received, 141.12 KiB sent       # essentially no traffic
                                                       # NO "latest handshake" line
```

`tcpdump` on vmss000000 confirmed inbound UDP from `157.58.218.4` (spark-2c24's NAT IP, working) and `157.58.218.71` (a stale spark-3d37 source from a previous handshake) but never from spark-3d37's *current* source IP. Spark was sending; vmss was either not receiving or filtering.

Comparing the two spark nodes' default routes, the cause was structural:

```
spark-2c24 (working):  default via 172.18.10.252 dev enP7s7    # apollo (lab gateway)
spark-3d37 (broken):   default via 172.18.10.254 dev enP7s7    # different router
```

`apollo` is the lab gateway with full NAT/MASQUERADE; `.254` is a non-NAT'ing path with a different egress IP and asymmetric return path. Spark-3d37 had been wrongly configured to use `.254` as default. The fix was a one-line netplan change to point the default at apollo.

But after switching the default route to apollo, the WG handshake **still didn't recover for over a minute**. Apollo's egress capture showed:

```
# apollo eno4 OUT (after spark default route change)
10.132.164.4.51821 -> 52.138.2.134.51820 length 148      # spark-2c24 NAT'd: src=apollo IP  OK
10.132.164.4.6714  -> 52.138.2.134.51820 length 148      # spark-2c24 NAT'd: src=apollo IP  OK
172.18.10.7.51821  -> 52.138.2.134.51820 length 148      # spark-3d37 NOT NAT'd!           BAD
```

Same destination, same MASQUERADE rule (`-A UNBOUNDED-MASQUERADE -o eno4 -j MASQUERADE`), same egress interface. Two flows from neighboring sparks. One NAT'd, one not.

Conntrack on apollo held the answer:

```
# apollo conntrack (BEFORE flush)
udp  src=172.18.10.7 dst=52.138.2.134 sport=51821 dport=51820  \
     src=52.138.2.134 dst=172.18.10.7 sport=51820 dport=51821 \
     [UNREPLIED] mark=0 use=1
```

Tuples are **mirror copies** — no SNAT was recorded. Manually flushing this single entry made the next packet take a fresh path through netfilter and pick up MASQUERADE correctly. After flush, conntrack showed:

```
# apollo conntrack (AFTER flush)
udp  src=172.18.10.7 dst=52.138.2.134 sport=51821 dport=51820  \
     src=52.138.2.134 dst=10.132.164.4 sport=51820 dport=63470 \
     [ASSURED]
```

Now `Reverse.DstIP=10.132.164.4` (apollo's external IP), `Reverse.DstPort=63470` (post-NAT port) — SNAT correctly applied, handshake completed within seconds, end-to-end probe went 20/20.

## Root cause

When a node has a Linux bridge interface on a downstream network and `net.bridge.bridge-nf-call-iptables=1` (required for k8s on the same node), every frame that transits the bridge invokes the netfilter hooks even when the node is not the routing target. This creates a hazard: if a downstream host's frames are bridged out the same VLAN (not routed via this node's default-gateway iface), conntrack records the flow as `UNREPLIED` with no NAT applied because the MASQUERADE rule is scoped to `-o <default-iface>`.

If the downstream host's default route later changes to point at this node, the same flow now SHOULD be SNAT'd. But Linux's NAT engine only computes NAT decisions on the **first packet** of a flow and reuses thereafter. Subsequent packets keep using the cached "no-NAT" decision. Packets leave with the wrong source address, the upstream peer never replies, and the affected flow stays broken until the conntrack entry ages out — which is effectively never for keepalived UDP like WireGuard, because every keepalive refreshes the entry's timeout.

This isn't a one-off bug — it's a structural hazard for any unbounded-net gateway node that bridges a downstream subnet (e.g., `apollo` bridging the lab spark VLAN). Any time a downstream host's routing topology changes, this class of staleness can recur.

## Fix

New helper in `internal/net/netlink/conntrack_flush.go`:

```go
func FlushStaleTransitUDPConntrack(localIPs []net.IP) (uint, error)
```

Scans the conntrack table and removes UDP entries that meet **all** of:

- protocol UDP
- original and reply tuples are exact mirrors (no SNAT/DNAT applied)
- neither tuple end is a local IP

The L2-transit-with-no-NAT case is precisely this shape. Legitimate node-originated or node-terminated flows are excluded by the local-IP check; legitimate NAT'd flows are excluded by the mirror-tuple check.

TCP is intentionally left alone — handshake state makes staleness less common, and forcibly killing a connected TCP session would be more disruptive than the self-recovery wait.

Wired into `syncMasqueradeRules` in `cmd/unbounded-net-node/site_watch_reconcile.go` so any reconcile (already periodic) opportunistically clears stale entries within seconds of a downstream route flip. No tunables, no new Site/Node configuration required.

## Tests

7 new unit tests in `internal/net/netlink/conntrack_flush_test.go` for the filter predicate:

- `TestStaleTransitUDPFilter_MatchesUntranslatedTransit` — the target case (mirror tuples, no NAT) matches
- `TestStaleTransitUDPFilter_SkipsNATTranslated` — NAT'd flows are preserved
- `TestStaleTransitUDPFilter_SkipsTCP` — TCP is not touched
- `TestStaleTransitUDPFilter_SkipsLocalSource` — node-originated flows are preserved
- `TestStaleTransitUDPFilter_SkipsLocalDestination` — node-terminated flows are preserved
- `TestStaleTransitUDPFilter_SkipsNilFlow` — defensive nil handling
- `TestStaleTransitUDPFilter_SkipsPortMismatch` — defensive: anything that isn't a clean mirror is skipped

## Verification on live cluster

Reproduced the failure on `vapa-unbounded-2`:
- spark-3d37 default route: `172.18.10.254` (no NAT) -> `172.18.10.252` (apollo) via netplan + `netplan apply`
- Apollo conntrack: 31 stale UDP entries pinned to spark-3d37; all matched the filter and were removed
- WireGuard handshake completed within seconds
- End-to-end probe: 0/10 -> 10/10 success on `curl http://ollama:11434/api/tags`

With this PR's daemon-side flush in place, the same recovery happens automatically on the next reconcile after any downstream route flip — no manual `conntrack -D` required.

## Risk

Low. The filter is conservative:

- requires exact mirror tuples (anything Linux NAT touched is preserved)
- excludes local-IP tuples (node-originated/terminated flows preserved)
- excludes TCP (no risk of resetting established connections)

Worst case: a transit UDP flow that legitimately had no NAT applied (e.g., a downstream host pinging another downstream host, transiting the bridge but not requiring egress) gets re-evaluated by netfilter on the next packet — same outcome as the existing path, just a one-RTT cost on the next packet.

phealy:
Not actually needed with proper ECMP working.

PR #116, opened 2026-05-04
docs: adding lifecycle operations

imiller31:


designs/lifecycle-operations.md
```diff
@@ -0,0 +1,81 @@
...
+operation ignore it.
+
+## Proposed operations
+
+| Operation | Boundary | Responsible component | Meaning |
+|---|---|---|---|
+| `HostPowerOff` | Host | `machine-ops-controller` for cloud VMs; `metalman` for PXE/bare metal with BMC | Power off the VM or physical host through the provider or BMC. |
+| `HostPowerOn` | Host | `machine-ops-controller` for cloud VMs; `metalman` for PXE/bare metal with BMC | Power on or start the VM or physical host through the provider or BMC. |
+| `HostReboot` | Host | `machine-ops-controller` for cloud VMs; `metalman` for PXE/bare metal with BMC | Reboot, reset, or power-cycle the host through the provider or BMC. |
+| `NodeStop` | Node container | `unbounded-agent` on the host | Stop kubelet and containerd inside the nspawn machine, then stop the nspawn machine. The rootfs remains intact. |
```
bcho: do we actually need `NodeStart` / `NodeStop`? Can we just use `NodeReboot` here?

imiller31: I can't really think of a case where you'd want the node down but also not the host  myself.

jwilder: Maybe useful for moving hardware to a different cluster or just bouncing kubelet to unwedge it from a bug?

designs/lifecycle-operations.md
```diff
@@ -0,0 +1,81 @@
...
+containerd, CNI plugins, and pod containers while leaving the host running.
+
+Operations are requested with `MachineOperation` and target a `Machine` by
+`spec.machineRef` or `spec.machineSelector`. Each operation is handled by the
+component that owns the relevant boundary. Controllers that do not own an
+operation ignore it.
+
+## Proposed operations
+
+| Operation | Boundary | Responsible component | Meaning |
```
jwilder, line 19: How is host OS image change handled?  Is that a specific Operation or a change to the desired state on the Machine and then triggered by one of thees operations (maybe HostReboot?)

designs/lifecycle-operations.md
```diff
@@ -0,0 +1,81 @@
...
+
+`MachineOperation` status follows a job-like lifecycle: `Pending`,
+`InProgress`, `Complete`, or `Failed`. Implementations should set
+`status.startedAt`, `status.completedAt`, a human-readable `status.message`, and
+conditions that identify the executor and failure reason. Completed and failed
+operations may be removed with `spec.ttlSecondsAfterFinished`.
+
+## Risks and open questions
+
+If no component claims an operation, the operation may remain pending forever.
```
jwilder, line 76: What happens if you call `HostPowerOn` on an SSH joined machine where it was provisioned and controlled externally?  Would it be clear somewhere that operation is not supported?  Would the machine-ops-controller just mark the operation as Failed with the reason?

docs/content/concepts/lifecycle-operations.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "Lifecycle Operations"
+weight: 5
+description: "Proposed operation boundaries and executor ownership for MachineOperation requests."
```
jwilder: Note: this document is under the public website docs so this will get published to the website when committed which I don't think is intended.  Maybe separately we should either move the website to a different dir or pub discussion docs like this elsewhere.

plombardi89: Bring back the Architectural Decision Record `adr/`? directory from one of the old repos we were working out of?

designs/lifecycle-operations.md
imiller31: This is a question @bcho and I had on how we want to handle. If we reimage a node then we need to have some script to install the agent or image baked with the agent. Do we want to support that?

designs/lifecycle-operations.md
imiller31: I think we need to define how unsupported operations are handled. What should the experience be for a user? We could have some sort of custom admission controller to handle the rejections, otherwise we likely need some fallback controller to clean up any unsupported ops

designs/lifecycle-operations.md
bcho: For the first one i think we can use reimage host, while the second one should be covered by node reboot. I prefer to skip node start/stop for now becuase it might make the state complicated when running alongside with host level start/stop.

designs/lifecycle-operations.md
bcho: if we reimage the host, we should rebuild/rebootstrap the node, this is very similar to how AKS RP handles  rolling node upgrade at this moment. So I assume we should be have a similar approach like doing ssh bootstrap through host? For instance, cloud-init if it's a vm

designs/lifecycle-operations.md
bcho: +1 on having admission controller/cel to block unsupported operations

bcho (dismissed):

designs/lifecycle-operations.md
jwilder: Sounds good.  Agree to keep surface areas smaller until we have clear need to expand it.

jwilder (changes requested):
I think just remove the NodeStart/NodeStop

imiller31:
> I think just remove the NodeStart/NodeStop

@copilot remove NodeStart and NodeStop operations. Also rename HostReimage to HostReplace to indicate the destructive nature of deleting and recreating a vm with cloud-init

jwilder (approved):

bcho (approved):

PR #120, opened 2026-05-05
Implement agent repave on node delete

bcho:
## Summary
- implement agent daemon repave when the local Node is deleted, resolving the desired MachineConfigurationVersion from the Machine spec
- apply MCV template fields to the desired agent config and record the applied configuration in Machine status
- introduce a daemon `nodeOperator` interface so reconcilers depend on explicit active-machine discovery, restart, and repave operations instead of host-mutating package functions
- add shared MCV naming helper, AgentConfig.DeepCopy, bootstrapper RBAC, and e2e coverage for a kubelet patch-version repave

## Validation
- `python3 -m py_compile hack/agent/e2e-kind/e2e.py`
- `go test ./cmd/agent/internal/daemon ./cmd/machina/machina/controller ./pkg/agent/config ./internal/provision ./pkg/agent/phases/rootfs`
- `git diff --check`


bcho:
@copilot run notice generation for me

imiller31 (approved):
LGTM, I think we need to have some sort of churn testing where we observe lots of repaves, but that's outside the scope of this PR

bcho:
> LGTM, I think we need to have some sort of churn testing where we observe lots of repaves, but that's outside the scope of this PR

yeah i think we need to setup soak and stree test for the current setup... 

@copilot please create an issue for this

PR #121, opened 2026-05-05
Add high level storage doc

jveski:
Rough draft to help align on the basics.


jveski:
I'll give this a closer read when I have more brain space in the morning. After skimming it I think the semantics are right, but it seems like there's a lot of coordination between processes that isn't necessary since all processes are targeting a shared filesystem or object store anyway. Maybe locking is necessary but I don't think we need sharding (the backing regional storage layer should effectively handle this for us).

jwilder (approved):
This makes sense to me.  Some clarifying ideas that came out of this (for me) are:

1) Standardize on S3 API on P2P cache miss from regional cache (regional cache maybe use NFS or other protocols internally)
2) P2P cache provides the interfaces (S3, filesystem, OCI registory, http proxy, etc..) for each node

PR #124, opened 2026-05-05
Orca - ORigin CAche

plombardi89:
# Orca origin cache: implementation, hardening, and dev tooling

## Summary

Adds the Orca origin cache as a complete v1: a horizontally-scaled
HTTP cache that fronts S3 / Azure Blob origins with chunked,
rendezvous-hashed, in-DC caching. Spans the design doc, the runtime
binary (`cmd/orca`), the in-process integration suite
(`internal/orca/inttest`), Kubernetes deployment manifests, and a
local dev harness with a Go-based seeding tool.

The PR is large because it combines:
1. The initial implementation of the cache.
2. Three rounds of review-driven hardening (about 35 commits of bug
   fixes, correctness tightening, and observability work).
3. The local dev harness (`hack/orca/`, `hack/cmd/orcaseed`) and its
   docs.

This description walks through what the system does, the major
components, and the high-leverage areas reviewers should focus on.

---

## What Orca is

A read-only origin cache. Clients speak a minimal S3-compatible
surface (`GET`, `HEAD`, `LIST`) to Orca; Orca fans out to a configured
origin (AWS S3 or Azure Blob Storage), splits objects into fixed-size
chunks (default 8 MiB), commits them to an in-DC S3-compatible
cachestore (atomic via `If-None-Match: *`), and serves subsequent
reads from cache.

A cluster of Orca replicas coordinates via **rendezvous-hashed
coordinator selection** plus a per-replica **singleflight**: at any
given time, exactly one replica is doing the origin GET for any given
chunk, and all concurrent client requests for that chunk collapse
onto it. The coordinator either fills locally or streams from a peer
via an internal RPC (`GET /internal/fill`).

No disk spool. No metadata DB. Chunk identity is encoded as a
deterministic SHA-256 over (origin_id, bucket, key, etag, chunk_size)
so the cachestore key set is itself the source of truth.

---

### Runtime packages

| Path | Responsibility |
|---|---|
| `internal/orca/server/` | HTTP edge (port 8443) + internal-fill listener (port 8444) + ops listener (port 8442 with `/healthz` and `/readyz`). |
| `internal/orca/fetch/` | The per-replica fetch coordinator: HeadObject metadata, GetChunk orchestration, singleflight, pre-header origin retry, fill streaming. |
| `internal/orca/cluster/` | Peer discovery (DNS-backed in prod, static in tests), rendezvous hashing, internal-RPC client. |
| `internal/orca/metadata/` | Per-replica metadata cache with TTL, negative-TTL, and singleflight HEAD. |
| `internal/orca/chunkcatalog/` | Presence-only LRU recording chunks known to be in the cachestore. |
| `internal/orca/cachestore/s3/` | S3-compatible cachestore driver (LocalStack in dev, VAST or similar in prod). Atomic commit via PutObject + If-None-Match. Boot-time versioning gate + self-test. |
| `internal/orca/origin/awss3/` | AWS S3 (and S3-compatible) origin driver. |
| `internal/orca/origin/azureblob/` | Azure Blob Storage origin driver (Block Blobs only). |
| `internal/orca/chunk/` | Chunk-key encoding (SHA-256 path computation, range-to-index math). |
| `internal/orca/config/` | YAML config + env-var fallbacks + `Logging.Level` + `ORCA_LOG_LEVEL` env override. |
| `internal/orca/app/` | Lifecycle wiring (Start / Wait / Shutdown), three HTTP listeners, dependency injection seams for tests. |

### Tests

- **Unit tests** in every package, covering driver branches, range
  math, error mapping, config defaults, manifest YAML validity.
- **Integration suite** (`internal/orca/inttest`, build tag
  `integrationtest`, target `make orca-inttest`): 7 tests against
  real LocalStack + Azurite containers via testcontainers-go, with
  N in-process orca instances wired to those backends. Covers
  cold/warm GET, cross-chunk ranges, 64-chunk multi-chunk, rendezvous
  routing, singleflight collapse with bounded origin-call counts,
  membership-disagreement 409 fallback, and azureblob end-to-end.

### Deployment

- `deploy/orca/*.yaml.tmpl` - production manifests (Namespace, RBAC,
  ConfigMap, Service x2 - headless peer Service + edge Service,
  Deployment with anti-affinity).
- `deploy/orca/dev/*.yaml.tmpl` - dev-only manifests (LocalStack,
  bucket-init Job, Azurite, container-init Job).
- `hack/cmd/render-manifests` - Go template renderer driven by
  repeatable `--set` flags.

### Dev harness

- `hack/orca/Makefile` - orchestration for a 4-node Kind cluster
  (1 control-plane + 3 workers) with extraPortMappings for NodePort
  30100 -> host loopback.
- `hack/orca/quickstart.md` - end-to-end recipe.
- `hack/orca/dev-harness.md` - long-form reference (origin modes,
  switching, troubleshooting).
- `hack/cmd/orcaseed/` - Go tool for populating the dev-cluster
  origin: `generate` (synthetic N x size random blobs with optional
  per-blob deterministic seeding), `upload` (file from disk),
  `list`, `delete`.

---

## Local dev-cluster workflow

For reviewers who want to bring this up themselves:

```bash
cp hack/orca/.env.example hack/orca/.env
$EDITOR hack/orca/.env  # set ORIGIN_DRIVER=azureblob LOG_LEVEL=debug
make orca-up
make -C hack/orca seed-generate SEED_ARGS='--size 10MiB --count 5'
make -C hack/orca port-forward  # in another terminal
curl -v http://localhost:8443/orca-test/synth-0 -o /dev/null
make -C hack/orca logs | jq 'select(.chunk.bucket=="orca-test")'
make orca-down
```

Full walkthrough in `hack/orca/quickstart.md`.


design/orca/design.md
```diff
@@ -0,0 +1,2735 @@
...
+    A->>B: GET /internal/fill?key=k2 (mTLS)
+    B-->>A: chunk bytes
+    A-->>C: stream slice
+    Note over A: k3: miss, Coordinator(k3) = D != self
+    A->>D: GET /internal/fill?key=k3 (mTLS)
+    D-->>A: chunk bytes
+    A-->>C: stream slice
+```
+
+### 8.4 Origin backpressure
```
jveski: How much do we care about getting throttled? I don't think coordinated rate ilmiting is worth the complexity compared to a simple per-process token bucket + exponential backoff

design/orca/design.md
```diff
@@ -0,0 +1,2735 @@
...
+  origin GET worst case). Sustained non-zero `failed` rate is a
+  CacheStore-health alert, not a per-request error path.
+- **Typed `CacheStore` errors during read**: `ErrNotFound` triggers the
+  miss-fill path; `ErrTransient` surfaces as `503 Slow Down` with
+  `Retry-After: 1s`; `ErrAuth` surfaces as `502 Bad Gateway`. Sustained
+  `ErrTransient` / `ErrAuth` trips the per-process **CacheStore circuit
+  breaker** (s10.2). Sustained `ErrAuth` (default 3 consecutive) flips
+  `/readyz` to NotReady so load balancers drain the replica.
+
+### 8.7 Metadata-layer singleflight
```
jveski: So if I'm reading this right the doc proposes consistent hashing of the file/blob keyspace across Orca replicas, plus in-memory pipelining to batch concurrent requests for the same blocks.

Any concerns about hotspots if we're only running a few replicas? Since we have shared storage under the cache it's feasible to avoid partitioning (at the cost of another network roundtrip for locking)

design/orca/design.md
```diff
@@ -0,0 +1,2735 @@
...
+  returns `ErrNotFound` and re-enters the miss-fill path.
+  Operators SHOULD NOT sweep the staging subdirectory
+  `<root>/.staging/` - that is managed by the driver's own
+  background sweep (`cachestore.posixfs.staging_max_age`, default
+  1h, s10.1.2).
+- **Active eviction** (s13.2): enable the cache layer's
+  access-frequency-driven eviction loop. This is the recommended
+  posixfs path when external sweep tooling is impractical.
+
+### 13.2 Active eviction (opt-in, access-frequency)
```
jveski: Do we definitely need eviction? I _assume_ the remote dataset might be larger than in-region storage, but worth confirming

design/orca/design.md
plombardi89: @jveski depends heavily on the number of replicas of the cache we think we would need. That's actually where I started and this was offered as an optimization.

I do prefer the simplicity of a token bucket and if we want to go forward with coordinated RL we at least have a design for a future implementation.

design/orca/design.md
plombardi89: I think @jwilder has mentioned that the origin data set is going to be bigger than the origin cache data store can store at least some of the times.

It feels like we need _some_ kind of eviction but maybe we need more clarity on the requirement

jveski: Cool, this approach checks out then. Being able to keep the counters in memory might justify sticking with consistent hashing even if it comes at the cost of some hotspotting across replicas. So I guess that mostly addresses both of my remaining comments. :upside_down_face: 

As long as we run fewer, "larger" replicas of this service I think we're good to go :+1:

design/orca/design.md
plombardi89: Reverted design to a token bucket and moved the coordinated rate limiting to a future optimization section

plombardi89:
@copilot+claude-opus-4.7 perform a review of this PR

internal/orca/fetch/fetch.go
```diff
@@ -0,0 +1,554 @@
...
+	)
+
+	if rc, hit, err := c.lookupOrStat(ctx, k, objectSize); err != nil {
+		return nil, err
+	} else if hit {
+		return rc, nil
+	}
+
+	// Cluster-wide dedup: route to coordinator.
+	coord := c.cl.Coordinator(k)
```
jveski, line 164: How does this perform for large blobs? A 700GB object would be ~90k 8MB chunks which from what I can tell will become 90k HTTP requests - seems like a lot of overhead

plombardi89: Good point, for a large blob like this you're going to end up with a lot of reads. I can look into adaptively scaling the chunk sizes with the size of the blob though bigger chunks mean more inflight memory usage and more time spent reading per chunk.

Open to other ideas if you have some.

jveski: Yeah I agree about just using larger chunks. Even multiple GB wouldn't be crazy with some clever buffering to avoid holding the entire chunk in memory.

jveski: One other thing to clarify: does this currently download the actual chunks from the peer process or just coordinate with it? Ideally we get the chunks from regional storage and use the Orca cluster's hash/rpc just for coordination.

plombardi89: Warm cache hit on the assembler. Today the assembler probes the chunk catalog plus regional storage itself first. If it hits, it streams the bytes from regional storage straight to the client. No peer call. This is the common path on a warm cluster. (Diagram 3 in design/orca/design.md.)

Cold miss. The chunk isn't in regional storage yet. The assembler asks the peer-coordinator via the internal RPC. The peer fetches from origin, holds the chunk in memory, streams it back to the assembler, and commits to regional storage in parallel (commit-after-serve). Here the peer is in the data path and the bytes only exist in the peer's memory until the commit lands. We accept this because it hides the commit cost behind the response stream. If we wanted I think we could get fancy here where we start the commit to regional storage and then respond to the peer to start streaming that chunk BUT it adds latency and some coordination as you're effectively telling the peer to go to a different system when the peer is already holding the bytes anyways.

Race-hit case. The assembler's probe missed (catalog miss + Stat returned NotFound), but by the time the peer's probe runs the chunk is in regional storage (committed by some other replica during the few-millis window). Today the peer streams those regional-storage bytes back to the assembler over the RPC, acting as a proxy.

The race case is small and easy to fix: have the peer respond "chunk is already in storage, go read it yourself" instead of proxying. The assembler then re-pokes regional storage and reads directly. It adds round-trips tho.

jveski: SGTM :ship:

jveski (approved):

PR #125, opened 2026-05-06
[machine-ops] Add host replace machine operation

imiller31:
## Summary
- Add `HostReplace` MachineOperation support with `kubectl unbounded machine replace` and destructive confirmation/`--force` handling.
- Implement Azure VM delete-and-recreate replacement using fresh cloud-init bootstrap while preserving VM name, NICs, data disks, identity, and managed disk settings where valid.
- Add controller bootstrap resolution, scoped RBAC, docs, and explicit unsupported handling for OCI host replacement.

## Testing
Tested locally against unbounded aks cluster with azure VMs

### DIFF
--- a/.github/workflows/agent-e2e-kind.yaml
+++ b/.github/workflows/agent-e2e-kind.yaml
@@ -210,6 +210,9 @@ jobs:
       - name: Validate workload on rejoined agent node
         run: python3 ./hack/agent/e2e-kind/e2e.py --verbose validate-workload
 
+      - name: Validate node repave upgrade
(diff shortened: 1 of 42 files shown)


bcho (commented):
can we use yaml library to construct the cloud-init file?

imiller31:
> can we use yaml library to construct the cloud-init file?

@copilot replace the string based cloud-init generation with a yaml library based one. Keep the code clean and maintainable.

bcho (approved):

PR #136, opened 2026-05-08
P2P cache protocol

jveski:
Adds a doc and simulator for the p2p protocol.

### DIFF
--- a/designs/p2p-protocol.md
+++ b/designs/p2p-protocol.md
@@ -1,0 +1,358 @@
+# P2P Storage Cache Protocol
+
+## Background
+
+A large cluster (think 100k nodes) needs to share content with itself:
+container image stripes, build artifacts, model weights. We want any
+node to fetch any object from any other node, with no central server.
+
+The goal is low latency and high aggregate throughput during a
+fleet-wide rollout - we want to saturate every node's local links,
+while keeping oversubscribed inter-rack uplinks lightly loaded.
+
+- Stripes are content-addressed and immutable: a key is the hash of its bytes
+- Ground truth for every object is **regional storage** external to the cluster
+- Objects are split into fixed **1 GiB stripes** and the protocol operates on one stripe at a time
+- The data path is **RDMA**, so a full mesh across all cluster nodes is not feasible
+
+## Protocol
+
+Node `C` wants stripe `K`. Nobody in the cluster has it yet.
+
+**Step 1: Find the stripe's owner.**
+
+Every node has the cluster membership list - the set of cache pods
+running across the fleet, learned from the Kubernetes API. Every node
+hashes `K` to a position on a ring of node IDs and reads off the owner:
+the node whose ID comes immediately after that position. Call it `O`.
+There is no lookup, no directory, no coordinator; every node computes
+the same answer locally.
+
+For redundancy/availability, the next `r-1` successors of `O` on the ring are also
+valid first-ask targets (default `r = 8`).
+
+```mermaid
+flowchart LR
+    K["hash(K)"] -. "first ID after" .-> O["O<br/>(owner)"]
+    O --> S1["successor 1"]
+    S1 --> S2["successor 2"]
+    S2 --> Sdots["..."]
+    Sdots --> Sr["successor r-1"]
+
+    classDef owner fill:#d4edda,stroke:#28a745,color:#000
+    classDef succ fill:#fff3cd,stroke:#ffc107,color:#000
+    class O owner
+    class S1,S2,Sdots,Sr succ
+```
+
+**Step 2: Pick a path toward the owner.**
+
+`C` does *not* talk to `O` directly. `O` could be anywhere - possibly
+across an oversubscribed spine link - and we want to keep traffic local
+whenever we can.
+
+Instead, `C` keeps two small **finger tables**, each sized to roughly
+the cube root of the cluster (a few thousand entries at 100k nodes):
+
+- a **level-2** set of peers chosen to evenly cover the ring, and
+- a **level-1** set under each level-2 peer, again covering the ring
+  beneath it.
+
+Together they let `C` reach any node in at most three hops. Entries
+are refreshed in the background by re-sampling the ring whenever
+membership changes; a dead finger is replaced by the next live ID in
+its slot.
+
+```mermaid
+flowchart TD
+    C["C (this node)"]
+    C --> L2["fingers_L2(K): V*b peers<br/>covering the ring"]
+    L2 --> H2a["H2 candidate"]
+    L2 --> H2b["H2 candidate"]
+    L2 --> H2dots["..."]
+    L2 --> H2n["H2 candidate (V*b total)"]
+    H2a --> L1["fingers_L1(K, H2): b peers<br/>covering the ring beneath H2"]
+    L1 --> H1a["H1 candidate"]
+    L1 --> H1dots["..."]
+    L1 --> H1b["H1 candidate (b total)"]
+    H1a --> O["O<br/>(at most one hop away)"]
+```
+
+Each peer carries a topology label (`rack/row/...`) read from its
+Kubernetes Node, so "nearest" is a local lookup. `C`
+picks its nearest level-2 peer (same rack, then same row, then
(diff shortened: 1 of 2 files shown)


designs/p2p-protocol.md
```diff
@@ -0,0 +1,358 @@
...
+- **Adequate steady-state throughput.** A BitTorrent-style
+  swarm or a tree multicast would converge a bit faster on a single
+  huge fanout, but both need a tracker. We use the routing fabric
+  we already have, pay a brief warm-up, and then ride the same rules
+  for both routing and caching.
+- **No special path for hot files.** No "broadcast" mode, no separate
+  tier-1 cache, no toggle. The same rules apply cold or hot. The
+  cost of that uniformity is that small workloads pay the three-hop
+  tax where a flat hash table would do one hop.
+- **No write/churn modeling.** The design assumes membership changes
```
vpatelsj, line 280: Wouldn't some of these scenarios cause churn?: 

1). GPUs go down constantly and if can't repaird without a reboot or reimage that would cause a node leave - join event in the ring.
2). Planned maintenances like OS upgrades, kube upgrades etc..
3). Unplanned downtime like power outage etc..

For the simulation, could we add a scenario where nodes constantly leave and join the cluster and measure how many pulls hit a stale (now-dead) finger, pulls that take more hops than designed because of fallback and cache effectiveness during high-churn windows.

designs/p2p-protocol.md
```diff
@@ -0,0 +1,358 @@
...
+`V*b + b^2` nodes per stripe (we'll define `V` and `b` shortly). A
+file split into `S` stripes therefore lives on roughly `S * (V*b + b^2)`
+serving slots - plenty for a fleet-wide fanout.
+
+This is a pull-through CDN, except it falls out of the routing rules.
+No broadcast. No tracker. No admin pre-positioning the hot files. A
+key gets popular and the system grows more serving slots for it.
+
+**What gets cached.** Every stripe that passes through is admitted to
+a single per-node LRU pool with a fixed byte budget (default 10% of
```
vpatelsj, line 146: Can we add some back of the envelope calculation about the cache capacity that we expect based on a 10% node disk budget?

vpatelsj, line 146: Here's LLM generated storage capacity estimates: 



---

## Storage capacity at 10k nodes — fanout vs unique content

I worked through the cache capacity math to sanity-check the design's sizing claims. There's a discrepancy between the doc's stated fanout formula and what the code implements, which materially affects the answer. Flagging both, plus the corrected numbers.

### The formula discrepancy

The doc states the per-stripe fanout ceiling as `V*b + b²`.

Reading `sample_placement` in the simulator:

```python
b_l2 = b * max(1, h2_virtual)            # V*b level-2 candidates
level2 = pool[:b_l2]
level1_per_h2: dict[int, np.ndarray] = {}
cursor = b_l2
for h2 in level2.tolist():
    level1_per_h2[h2] = pool[cursor:cursor + b]   # b level-1 candidates per H2
    cursor += b
```

The code creates `V*b` H2 candidates, and for *each* H2, `b` L1 candidates beneath it. Total L1 cache slots: `V*b × b = V*b²`, not `b²`.

So the implemented ceiling is:

`V*b + V*b² = V*b·(1+b)` ≈ `V*b²` for large b

**Could the maintainer confirm:** is the doc's formula a typo (should be `V*b + V*b²`), or is the code wrong (L1 set should be shared across H2s and total `b²`, not `V*b²`)? The choice affects capacity planning by a factor of V.

The numbers below assume the code is correct (`V*b·(1+b)` ceiling). If the doc's formula is the intended design, divide the "full fanout" content figures by V.

### Parameters at 10k nodes

- `N = 10,000`
- `b = ⌈N^(1/3)⌉ = 22`
- `V = 2` (default)
- Stripe size = 1 GiB
- Full-fanout ceiling: `V*b*(1+b) = 2 × 22 × 23 = 1,012` cache slots per stripe
- Half-fanout: ~506 cache slots per stripe
- Quarter-fanout: ~253 cache slots per stripe

### Unique content the cluster can cache, by disk size

These numbers assume the cache is filled exclusively with stripes at the indicated fanout level — nothing else competing.

| Disk per node | Per-node cache (10%) | Total cluster cache | Full fanout (~1,012 copies) | Half fanout (~506 copies) | Quarter fanout (~253 copies) |
|---|---|---|---|---|---|
| 500 GB | 50 GB | 500 TB | **~494 GB** unique | **~988 GB** unique | **~1.93 TB** unique |
| 1 TB | 100 GB | 1 PB | **~988 GB** unique | **~1.93 TB** unique | **~3.86 TB** unique |
| 2 TB | 200 GB | 2 PB | **~1.93 TB** unique | **~3.86 TB** unique | **~7.72 TB** unique |
| 4 TB | 400 GB | 4 PB | **~3.86 TB** unique | **~7.72 TB** unique | **~15.4 TB** unique |
| 10 TB | 1 TB | 10 PB | **~9.66 TB** unique | **~19.3 TB** unique | **~38.6 TB** unique |

### What "fanout" means for performance

- **Full fanout (~1,012 copies per stripe):** the regime where the design's claimed properties hold — owner is idle, most reads are one hop, almost always intra-rack.
- **Half fanout (~506 copies):** still highly redundant, performance should be close to the design's claims with slightly more route variance.
- **Quarter fanout (~253 copies):** the redundancy structure starts to degrade. More requests will route 2-3 hops; cross-rack share rises; some load returns to owners. Still usable but no longer matches the doc's headline claims.
- Below quarter fanout: degraded regime; the design is operating outside its happy path.

### Workload sizing — concrete examples at 1 TB disks

With 1 TB disks (~988 GB of full-fanout unique content):

**Container images** (assume ~500 MB average):
- Full fanout: ~1,975 images cached fleet-wide ✓ plenty
- Half fanout: ~3,950 images
- Quarter fanout: ~7,900 images

**AI models — small (e.g., Llama-3 8B quantized, ~7 GB):**
- Full fanout: ~141 models ✓ comfortable
- Half fanout: ~282 models

**AI models — medium (e.g., Llama-3 70B quantized, ~140 GB):**
- Full fanout: ~7 models ⚠ tight
- Half fanout: ~14 models
- Quarter fanout: ~28 models

**AI models — large (e.g., 405B fp16, ~700 GB):**
- Full fanout: 1 model ✗ insufficient at 10% budget
- Half fanout: ~2 models
- Quarter fanout: ~5 models

vpatelsj (changes requested):
My concerns: 

1). Forward and Cache only pays off when cache size  is >>> working set. At 10k nodes with 1 TB disks at 10% budget, full-fanout capacity is ~1 TB. One 700 GB model fills it. Every subsequent pulls that does warming essentially keeps causing evictions constantly. 

2). custom ring discovery is very clever and novel idea. But if we are doing most of the dataplane over RDMA, the discovery hops are very cheap. A kad-dht that provides log(n) hops vs 3 hops on our custom ring doesnt provide much value and we take upon implementing something custom vs using something that is quite matured over the years.

designs/p2p-protocol.md
jveski: The agent did find a valid mismatch of terms in the doc - well kind of just a simplification but still worth fixing. But its capacity numbers are wrong. Will address in the your comment.

jveski: It's dependent on how hot a particular chunk is. The effective min copies of a chunk is 2, max is determined by the formula in the doc - roughly 1000 nodes on a 10,000 node cluster.

So for workloads with more, cooler blobs I'd expect 70-80% of the aggregate disk pool to be usable. Worst case with very hot keys more like 10-20%.

designs/p2p-protocol.md
jveski: Clients will naturally agree on the same fallback chain for the lost node, minus some slop to account for the topology biasing of the ring. So _at worst_ the cost of losing a node is a client-side timeout and a cold pull through the fallback.

Timeouts will amortize nicely with circuit breakers. Cold pulls could technically form a live lock if enough nodes churn in/out of the cluster around the same time. But that's also true of every distributed system: keep ripping the server away from the client and the client will by definition never succeed. :upside_down_face: That said- the shorter neighbor tree (compared to the standard Chord ring) will help here: less hops == less risk of being impacted by a degraded node.

PR #153, opened 2026-05-14
Refactor daemon controller setup

bcho:
## Summary
- Add reusable `pkg/agent/daemon` controller building blocks for host-local daemon work, including serialized `daemon.Request` dispatch for machine operation and repave requests.
- Move Machina `MachineOperation` watch mapping, local-target filtering, handler dispatch, unsupported-operation failure, and status persistence into `MachinaMachineOperationReconciler`.
- Update the Unbounded agent daemon to compose the reusable machine operation reconciler with separate Unbounded-owned operation handlers and repave reconciler.
- Add `NoopMachineOperationReconciler` for external providers that only need repave-style triggers and should fail unexpected machine operation requests.
- Document the daemon controller design, external provider adoption shape, and generic plus Unbounded-specific request/status flows with diagrams.

## Testing
- `go test ./cmd/agent/internal/daemon ./pkg/agent/daemon`
- `git diff --check`


pkg/agent/daemon/machineoperation.go
```diff
@@ -0,0 +1,242 @@
+// Copyright (c) Microsoft Corporation.
```
imiller31, line 1: After looking through this package, I don't see much that is specific to host-local daemon operations besides the repave reconciler. Is the intent to converge all MachineOperations controllers (external and host-local) to use a similar setup (i.e. this generic implementation)?

imiller31 (dismissed):

pkg/agent/daemon/machineoperation.go
bcho: this is more like a wording issue! Let me polish this a bit

bcho: updated the wordings

imiller31 (approved):

PR #157, opened 2026-05-18
net-node: make all tunnel-device names configurable end to end

phealy:
The unbounded-net-node agent has historically used hardcoded names for every tunnel device it creates: `geneve0`, `vxlan0`, `ipip0`, and the per-port WireGuard devices `wg<port>`. `--geneve-interface` already existed but its renaming aspect was incomplete (status reporting, peer-interface mapping, cleanup loops, and route annotations all keyed off the hardcoded default). VXLAN, IPIP, and the WireGuard prefix had no knob at all.

This PR finishes the job: every tunnel-device name the agent owns is now configurable via CLI flag and YAML, plumbed through every consumer, and validated at startup. Motivating use case: coexistence with another local CNI on the same hosts (e.g. a site running Cilium alongside unbounded-net) where the operator wants non-colliding device names.

## Configuration surface

| Flag | YAML | Default |
|---|---|---|
| `--geneve-interface` | `geneveInterfaceName` | `geneve0` |
| `--vxlan-interface` | `vxlanInterfaceName` | `vxlan0` |
| `--ipip-interface` | `ipipInterfaceName` | `ipip0` |
| `--wireguard-interface-prefix` | `wireGuardInterfacePrefix` | `wg` |

Per-port WireGuard devices are named `<prefix><port>` (default `wg51820` etc).

Startup validation:

* Shared tunnel names: non-empty; ≤15 bytes (IFNAMSIZ); pairwise distinct; must not equal `unbounded0`.
* WireGuard prefix: non-empty; ≤10 bytes (so `prefix + 5-digit port` fits in IFNAMSIZ); no `/`; must not equal `unbounded0`.

The dedicated `unbounded0` dummy device is deliberately not configurable.

## Plumbing

Every site that read from a hardcoded name now reads from `cfg`:

- `tunnel_config.go`: `configureTunnelPeers`, `cleanupUnusedTunnelDevices`, `reapplyRPFilterOnActiveTunnels`, `peerIfaceName`. New helpers `wireGuardInterfaceName(cfg, port)` and `isWireGuardInterface(cfg, name)`.
- `status_server.go`: `tunnelInterfaceName`/`isManagedTunnelInterface` take `cfg`.
- `route_annotations.go`: `markSupernetRoutesExpected`, `annotateFamilyRoutes`, `annotateRoutePeerDestinationsForFamily`, `isConnectedSelfWireGuardHostRoute`. The legacy `isWireGuardInterfaceName` / `isTunnelInterfaceName` helpers are replaced with a cfg-aware `isPeerRoutingInterface`.
- `internal/net/netlink/route_manager.go`: `NewUnifiedRouteManager` gains a `wgPrefix` argument (used by the orphan-route cleanup scan).
- `internal/net/routeplan/routeplan.go`: `BuildExpectedWireGuardRoutes` takes a new `InterfaceNames` struct (WG prefix + the three shared names); the dead routeplan-internal helper is removed.

## MAC at creation time (latent race fix)

End-to-end validation surfaced a long-standing race in `LinkManager`: `LinkAdd` returns before the kernel finishes async MAC assignment on flow-based GENEVE/VXLAN devices, and the subsequent `LinkSetHardwareAddr` is silently dropped on the race window, leaving the device with a kernel-random MAC. Remote agents derive inner-frame destination MACs from peers' underlay IPs via a deterministic formula (\`02:<ip-octets>:ff\`), so any remote talker would inner-MAC-mismatch us and the VXLAN/GENEVE driver would drop the frames.

Fixed by extending `EnsureGeneveInterface` and `EnsureVXLANInterface` to accept a `mac net.HardwareAddr` parameter and pass it through `LinkAttrs.HardwareAddr` at `LinkAdd` time. The post-creation `SetLinkAddress` call is retained as a safety net for the existing-device path. IPIP is unaffected (L3, no MAC).

## Stale-device cleanup (scoped narrowly to coexist with other CNIs)

Two new helpers handle rename transitions without touching other tooling:

- **`cleanupStaleFlowBasedTunnels`** (tunnel_config.go): deletes any flow-based GENEVE/VXLAN device whose **UDP port** matches `cfg.GenevePort`/`cfg.VXLANPort` AND whose name does not match the configured name. Matching on the configured port keeps the sweep safe when another CNI runs flow-based encap on a different port. Two agents on the same host that pick the same port can't coexist regardless — only one UDP listener per port.

  **IPIP is intentionally not swept** here; the kernel IPIP external-mode device exposes no field that distinguishes ours from another agent's. Operators renaming `--ipip-interface` must `ip link del <old-name>` manually (documented).

- **`removeStaleWireGuardInterfaces`** (reconciliation_helpers.go): uses `wgctrl` to query each `wireguard` kernel device's `ListenPort`, then deletes a device only when its current name matches the strict pattern `<alpha-prefix><exact-listen-port>` AND that prefix differs from `cfg.WireGuardInterfacePrefix`. Tooling that uses a different naming convention (`tailscale0`, plain `wg0`, UUID-named devices, …) is left alone.

## Validation

**Unit tests**:
- New `tunnel_interface_validation_test.go` with 20 subtests covering the happy path and every rejection case for both validators.
- Existing config/status/routeplan tests updated for the new defaults; `go test ./cmd/unbounded-net-node/... ./internal/net/... -count=1` passes.

**Dev cluster** (s1-aks-admin, 22 nodes across three sites):

| Scenario | Result |
|---|---|
| Defaults | 22/22 fresh, 0 problems, 16/16 VXLAN MACs match the deterministic formula, goldpinger 22/22 OK |
| `--vxlan-interface=vxa --wireguard-interface-prefix=uw` | Stale-cleanup logs show targeted removal of `vxlan0` and `wg<port>` devices; new `vxa` + `uw<port>` devices created with correct MACs; 22/22 fresh, 22/22 goldpinger OK |
| Restored defaults | Stale `vxa` + `uw*` swept, fresh `vxlan0` + `wg*` created; 22/22 fresh, 0 problems, 22/22 goldpinger OK |

## Diff stat

`21 files changed, 816 insertions(+), 195 deletions(-)`

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


cmd/unbounded-net-node/reconciliation_helpers.go
```diff
@@ -658,3 +664,102 @@ func removeUnmanagedWireGuardInterfaces(cfg *config, state *wireGuardState, desi
...
+// tooling typically names devices "wg0" or "tailscale0" or a UUID, which
+// won't match the strict "<alpha-prefix><exact-listen-port>" pattern.
+//
+// If an operator's other tooling happens to use the same "<alpha><port>"
+// pattern on the SAME UDP port, they cannot coexist on the same host
+// regardless of this code (only one UDP listener per port).
+func removeStaleWireGuardInterfaces(cfg *config) {
+	wgClient, err := wgctrl.New()
+	if err != nil {
+		klog.V(2).Infof("WireGuard: failed to open wgctrl for stale-prefix sweep: %v", err)
```
plombardi89: Info logs for errors? Is there an errorf or at least warnf func we could use?

internal/net/netlink/route_manager.go
```diff
@@ -114,15 +119,23 @@ type UnifiedRouteManager struct {
...
+// kernel routes pointing at managed WireGuard devices; pass "" to default
+// to "wg".
+func NewUnifiedRouteManager(linkName string, defaultTable int, wgPrefix string) *UnifiedRouteManager {
 	effectiveTable := defaultTable
 	if effectiveTable == 0 {
 		effectiveTable = unix.RT_TABLE_MAIN
 	}
 
+	if wgPrefix == "" {
+		wgPrefix = "wg"
```
plombardi89: It would be good to somehow get this "wg" as a const that both here and elsewhere such as https://github.com/Azure/unbounded/pull/157/changes#diff-4e73dabd57c0dd7d34f4075fb7a8345f58912e743bd61288366a53fb781b5a06R288 could reference the prefix and we don't have potential for one area changing and the other area not.

plombardi89 (changes requested):
Noticed a couple minor things

internal/net/netlink/route_manager.go
phealy: Good catch. Fixed.

cmd/unbounded-net-node/reconciliation_helpers.go
phealy: Yeah, that should be warnings (they prevent cleanup of an old interface but don't break functionality). Fixed.

plombardi89 (approved):

PR #160, opened 2026-05-18
[wip] Refactor MachineOperation controllers

imiller31:
## Summary
- add shared MachineOperation controller pattern with target resolution, operation registry, lifecycle store, and TTL cleanup
- refactor agent and cloud controllers onto it
- keep cloud provider logic internal

## Tests
- go test ./pkg/machineops/controller ./pkg/agent/daemon ./internal/machineops ./internal/machineops/providers/azurevm ./internal/machineops/providers/ociinstance ./cmd/agent/internal/daemon ./cmd/machine-ops-controller

### DIFF
--- a/pkg/agent/daemon/types.go
+++ b/pkg/agent/daemon/types.go
@@ -11,49 +11,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 
 	machinav1alpha3 "github.com/Azure/unbounded/api/machina/v1alpha3"
+	machineopscontroller "github.com/Azure/unbounded/pkg/machineops/controller"
 )
 
 // MachineOperation is a typed host-local daemon operation request.
-type MachineOperation struct {
-	// Name is the operation name in its source system.
-	Name string
-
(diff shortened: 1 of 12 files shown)


pkg/agent/daemon/types.go
```diff
@@ -11,49 +11,23 @@ import (
...
-	// MarkInProgress records that op has started execution with message as the
-	// human-readable in-progress status.
-	MarkInProgress(context.Context, MachineOperation, string) error
-
-	// Finish records the terminal or non-terminal result for op. The result
-	// carries the desired phase, reason, message, observed generation, and any
-	// controller-runtime requeue request.
-	Finish(context.Context, MachineOperation, MachineOperationResult[TGeneration]) error
-}
+type MachineOperationStore = machineopscontroller.Store
```
bcho, line 30: I actually need this to use different generation type because in aks flex node scenario the operation store is arm resource and not using int64

imiller31:
shelving this PR for now, seems like abstraction too early

PR #162, opened 2026-05-18
Add unbounded-storage blockstore

jveski:
Adds most of the core storage layer components required by the P2P cache.

- Store hugepages-aligned chunks on raw block devices
- Very high throughput (topology aware pinned threads w/ iouring)
- SIEVE(ish) eviction
- Admission bloom filter
- Never fsyncs for the sake of perf, still guarantees no torn pages (see tla+ model)


cmd/unbounded-storage/src/storage/admission/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Increment the sketch without consulting the doorkeeper.
+    /// Exposed so the engine can record hits as well as admits.
+    pub fn record_frequency(&self, key: &PageKey) {
+        let mut inner = self.inner.lock().unwrap();
+        sketch_bump(&mut inner.sketch, self.sketch_width, key);
+    }
+}
+
+fn doorkeeper_probe_and_set(words: &mut [u64], bits: u64, key: &PageKey) -> bool {
+    let h = key.mix(0);
```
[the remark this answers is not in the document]
jveski: This hash is only used for sharding so there are no security implications here. Pushed a commit to rename things to make that more clear.

cmd/unbounded-storage/src/storage/singleflight.rs
```diff
@@ -0,0 +1,259 @@
...
+impl Singleflight {
+    pub fn new(shards: usize) -> Self {
+        let shards = shards.max(1);
+        Self {
+            shards: (0..shards).map(|_| Mutex::new(HashMap::new())).collect(),
+        }
+    }
+
+    fn shard(&self, key: &PageKey) -> &Mutex<HashMap<PageKey, Arc<Lease>>> {
+        let h = key.mix(0);
```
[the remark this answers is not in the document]
jveski: This hash is only used for sharding so there are no security implications here. Pushed a commit to rename things to make that more clear.

cmd/unbounded-storage/models/.gitignore
```diff
@@ -0,0 +1 @@
+states/
```
cchildress, line 1: Should this just be in the root level .gitignore?

jveski: I'm fine with merging into the root .gitignore but I slightly prefer separate ones for cases where the directory the rule applies to might move or be deleted

cchildress: I don't have a strong opinion. This just caught my eye as something an LLM might have auto-generated in a stray directory without getting spotted.

cchildress (approved):

PR #166, opened 2026-05-19
feat: add site-scoped machine operation credentials

imiller31:
## Summary
- Add `MachineOperationCredential` for site-scoped MachineOperation auth configuration.
- Resolve provider auth from the Machine site label and pass it through the MachineOps provider interface.
- Replace provider-specific auth enums with two credential modes: `WorkloadIdentity` and `ExternalPlugin`.
- Keep credential Secret reads scoped to the machine-ops controller namespace.
- Remove legacy OCI startup auth flags and config Secret mounts.

Manual live testing was performed on the AKS E2E cluster with non-disruptive `HostPowerOn` probes for Azure and OCI site credentials.


api/machina/v1alpha3/machineoperationcredential_types.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Provider identifies the external control provider this credential applies
+	// to.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=AzureVM;OCIInstance
+	Provider string `json:"provider"`
+
+	// AuthType identifies how the provider should authenticate.
+	// +kubebuilder:validation:Required
+	AuthType MachineOperationAuthType `json:"authType"`
```
bcho: i am wondering would it be easier to only support two modes: 

- wokload identity / oidc auth via service account token (for provider supports this)
- external plugin (for provider doesn't support WI based auth)

then we don't need to maintain a long list of MachineOperationAuthType enums

imiller31: I think that makes sense, but I do have questions on how the UX will feel for that pattern.

cmd/agent/internal/daemon/register_test.go
```diff
@@ -70,6 +70,18 @@ func Test_buildMachineCR_NilLabelsAndTaints(t *testing.T) {
 	assert.Nil(t, machine.Spec.Kubernetes.RegisterWithTaints)
 }
 
+func Test_buildMachineCR_SiteLabelFromKubeletLabels(t *testing.T) {
```
plombardi89: Doesn't seem like this is exhaustive. Do we need a test for the more general `unbounded-cloud.io/site` label?

cmd/kubectl-unbounded/app/machine_register.go
```diff
@@ -208,6 +208,9 @@ func (h *machineRegisterHandler) executeAfterValidation(ctx context.Context) err
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name: h.name,
+			Labels: map[string]string{
+				"unbounded-cloud.io/site": h.siteName,
```
plombardi89: We should introduce a `const` for this label somewhere... we seem to reference it a lot.

cmd/machine-ops-controller/main.go
```diff
@@ -103,15 +99,21 @@ func run(ctx context.Context, cfg config) error {
 		return fmt.Errorf("create direct client: %w", err)
 	}
 
+	credentialSecretNamespace := cfg.credentialSecretNamespace
```
plombardi89: Is this intended? It's not documented in the command line help above that the default is to inherit the leader election namespace... I would not expect that as a user.

internal/machineops/auth.go
```diff
@@ -0,0 +1,168 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	unboundedv1alpha3 "github.com/Azure/unbounded/api/machina/v1alpha3"
+)
+
+const (
+	machineSiteLabelKey = "unbounded-cloud.io/site"
```
plombardi89: Here are these labels again. We should promote these to a more central place.

internal/machineops/auth.go
```diff
@@ -0,0 +1,168 @@
...
+		return "", fmt.Errorf("auth secret data key %q is required", key)
+	}
+
+	return value, nil
+}
+
+func (r *MachineOperationReconciler) resolveOperationAuth(ctx context.Context, machine *unboundedv1alpha3.Machine) (*OperationAuth, *authResolutionFailure, error) {
+	target, ok := operationAuthTargetFor(machine)
+	if !ok {
+		return nil, nil, nil
```
plombardi89: No `errors.New` is intended here on the `!ok` path?

plombardi89 (changes requested):
Small things to look at. I would also scan the various returned errors and look to see if they should be wrapped and annotated with `fmt.Errorf`

api/machina/v1alpha3/machineoperationcredential_types.go
bcho: we can take a look at crossplane. For external plugin case, I assume we want to ask user to mount a binary into the controller pod and they can configure how the binary retrieves cred via other settings. The only concern is we need to explicitly call out they should only put trusted binary here as the controller has no way to validate it

imiller31: I think in a production deployment one would deploy a controller per site since AFIAK you can only have a single workload identity per pod. 

Maybe we just have the extensibility for external auth for now and require a use to roll it themselves.

cmd/agent/internal/daemon/register_test.go
imiller31: expanded this

cmd/kubectl-unbounded/app/machine_register.go
imiller31: extracted this to a const, good catch!

cmd/machine-ops-controller/main.go
imiller31: I moved the default into the command and aligned with the leader election namespace. These will likely change in a subsequent PR as I crack this controller into multiple smaller ones scoped to the specific providers

internal/machineops/auth.go
imiller31: oh nice catch, I think this was some attempt to keep backwards compatibility, updated it

plombardi89 (approved):

PR #176, opened 2026-05-21
orcadev tool (benchmarking, dev setup, etc.)

plombardi89:
## Why this PR exists

Other developers are starting to build components that talk to Orca. They need a simple way to get Orca running in a Kubernetes cluster (kind on their laptop or an existing AKS / EKS / k3d cluster), put some test data into it, and run benchmarks or canned scenarios against it. Before this PR the answer was "read three READMEs, copy an env file, edit it, run several make targets across two directories, and hope the kind NodePorts work." That is not a coherent entrypoint.

## What this PR adds

Three things, in roughly increasing order of footprint:

**1. A developer tool: `bin/orcadev`** (in `hack/cmd/orcadev/`)

One command-line tool for everything a developer wants to do against a running Orca:

| Command | What it does |
|---|---|
| `orcadev upload` | Put a file (or N random blobs) into the origin. |
| `orcadev list` | Show what's in the origin. |
| `orcadev delete` | Remove things from the origin. |
| `orcadev roundtrip` | Upload a file, fetch it back through Orca, check the bytes match. |
| `orcadev cache list / inspect / clear` | See or remove what's in Orca's cache. |
| `orcadev bench` | Run a parallel-GET benchmark. Outputs human text and JSON for comparing runs. |
| `orcadev scenario` | Run a canned end-to-end scenario (`cold-warm`, `range-stress`, `empty-object`, `etag-change`). |

The tool figures out how to reach the cluster automatically: it opens `kubectl port-forward` connections to the three services it needs (the Orca edge, Azurite for origin storage, LocalStack for the cache) and tears them down when the command finishes. The same commands work on kind and on any other cluster you can reach with `kubectl`.

**2. A single install script: `./hack/orca/setup-orca.sh`**

One script with a small number of flags. It builds the Orca manifests on the fly, applies them to whatever kubectl context you point it at, waits for everything to be ready, and prints the next-step commands. The default install runs:

- Orca itself (3 replicas).
- Azurite as the Azure Blob Storage origin.
- LocalStack as the S3 cache store.
- Zero real cloud credentials required.

Two helper scripts handle the kind cluster lifecycle: `kind-up.sh` (create a kind cluster suitable for Orca) and `kind-down.sh` (delete it). The root Makefile exposes `make orca-kind-up`, `make orca-kind-down`, `make orca-install`, and `make orca-reset` as muscle-memory wrappers.

**3. One quickstart README: `hack/orca/README.md`**

A single document that walks a developer from "I have nothing" to "I am running benchmarks." It replaces the previous mix of `quickstart.md`, `dev-harness.md`, and a `.env.example` file (all deleted in this PR).

## What reviewers should look at first

In order of importance:

1. **`hack/orca/README.md`**. This is what new developers will read. If it doesn't make sense to a colleague who has never touched Orca, the PR has failed at its primary goal. Read it top to bottom and try the commands.

2. **`hack/orca/setup-orca.sh`**. Especially the uninstall path (label-based delete + the `--delete-namespace` opt-in) and the safety rails on `--build` / `--kind-load`. This script is the one thing developers will run on their clusters; mistakes here can delete unrelated resources.

3. **`hack/cmd/orcadev/orcadev/portforward.go` and `port_forward_coverage_test.go`**. The port-forward logic is what makes `bin/orcadev` work the same on kind and non-kind clusters. The coverage test asserts that every subcommand that talks to a storage service opens the right forwards first; please confirm this contract feels right.

4. **`deploy/orca/04-deployment.yaml.tmpl`** and **`deploy/orca/dev/*.yaml.tmpl`**. Two template changes worth a careful read: the new `RequireAntiAffinity` knob in the Orca Deployment (kind keeps `required`, others get `preferred`), and the switch from NodePort to ClusterIP for Azurite and LocalStack.

5. **`Makefile`** changes to the `##@ Orca` block. New / renamed targets: `orca-install`, `orca-kind-up`, `orca-kind-down`, `orca-reset`. Old aliases `orca-up` / `orca-down` are kept so existing muscle memory keeps working.

6. **`designs/orca/dev-workflow-remediation-plan.md`**. The plan document captured during the second review round; useful context for "why is this structured this way."

Files you can probably skim:
- All `*_test.go` additions: structural guards, no production behavior.
- `hack/cmd/orcadev/orcadev/*.go` outside `portforward.go` / `preset.go`: the subcommand bodies are mostly the original orcadev work that was already on this branch before the review.

## How to try it

```bash
# Build the tools.
make orca-build orcadev

# Bring up kind + install Orca + Azurite + LocalStack.
make orca-kind-up

# Verify.
kubectl --context kind-orca-dev -n unbounded-kube get pods

# Seed and exercise.
bin/orcadev upload --generate --count 5 --size 10MiB
bin/orcadev roundtrip --file /tmp/some-file.bin
bin/orcadev scenario cold-warm
bin/orcadev bench --key synth1 --duration 30s

# Tear down.
make orca-kind-down
```


hack/cmd/orcadev/orcadev/edge.go
```diff
@@ -0,0 +1,113 @@
...
+	Header http.Header
+	Body   io.ReadCloser
+	Size   int64
+	ETag   string
+}
+
+func newEdgeClient(baseURL string, timeout time.Duration) *edgeClient {
+	return &edgeClient{
+		baseURL: strings.TrimRight(baseURL, "/"),
+		http: &http.Client{
```
jveski, line 68: nit: probably want to define a roundtripper with higher max conns per host than the default

jveski (dismissed):
LGTM, one nit

hack/cmd/orcadev/orcadev/edge.go
plombardi89: Good catch - fixed in 7b92e1a. Cloned http.DefaultTransport, raised MaxIdleConnsPerHost to 256 (covers the README's --concurrency 16 plus any realistic dev workload without thinking about transport tuning), and pinned MaxConnsPerHost=0 explicitly so the operator's --concurrency choice is never silently throttled. Added TestNewEdgeClientTransportTuning to lock both values so a future refactor cannot regress them.

jveski (approved):

PR #186, opened 2026-05-27
unbounded-storage main.rs config plumbing

jveski:
Replaces dynamic discovery of block devices with an explicit config file to avoid potentially writing over mounted devices, and also models the peer connections in the same config file. Switches to Clap for the CLI instead of hand rolled arg passing.


plombardi89:
**Heads Up: Experiment in AI Code Review**

Commit 84685f0 ("replay every backing on topology generation change") replaced the old `Arc<LocalStorage>` pointer comparison in `LiveShardLocalStore` with a comparison against `LiveDiskTopology::generation()`. The commit message correctly calls out the bug in the old code: `registered` and `last_seen_ptr` lived under two separate mutexes, so a concurrent `apply_engines` could slip in between the two locks inside `register_backing`. Folding both fields into one `ReplayState` mutex fixes that bug.

But the second half of the change - swapping the pointer marker for a generation counter - trades one bug for another. The new check in [`shard_view.rs:103-121`](https://github.com/Azure/unbounded/blob/971f7ccd610a0aef8df6569c2d684a64705b1aa4/cmd/unbounded-storage/src/disk_supervisor/shard_view.rs#L103-L121) is:

```rust
let ls = self.topology.current()?;                  // (A) load snapshot
let gen_n = self.topology.generation();             // (B) load generation
let mut guard = self.state.lock().unwrap();
if guard.last_seen_generation != Some(gen_n) {
    Self::replay_locked(&ls, &guard.registered);    // replay against ls from (A)
    guard.last_seen_generation = Some(gen_n);       // record gen from (B)
}
Some(ls)
```

And the writer in [`topology.rs:91-93`](https://github.com/Azure/unbounded/blob/971f7ccd610a0aef8df6569c2d684a64705b1aa4/cmd/unbounded-storage/src/disk_supervisor/topology.rs#L91-L93):

```rust
self.current.store(Arc::new(LocalStorageSnapshotInner(snapshot)));   // (W1) store snapshot
let gen_n = self.generation.fetch_add(1, Ordering::AcqRel) + 1;      // (W2) bump generation
```

The comment at [`shard_view.rs:105-113`](https://github.com/Azure/unbounded/blob/971f7ccd610a0aef8df6569c2d684a64705b1aa4/cmd/unbounded-storage/src/disk_supervisor/shard_view.rs#L105-L113) says "we can never observe a new gen paired with an old snapshot". That is not what the memory model guarantees. The two atomics are independent. The reader's load at (A) and its load at (B) are separate operations with no ordering relationship between them, so the reader can see "old snapshot, new generation".

### The race

Here is one interleaving that breaks the check:

| step | thread | action | result |
|------|--------|--------|--------|
| 1 | shard | (A) `current.load_full()`   | sees S1 (writer has not stored yet) |
| 2 | main  | (W1) `current.store(S2)`    | S2 is now published |
| 3 | main  | (W2) `generation.fetch_add` | G is now 2 |
| 4 | shard | (B) `generation.load`       | sees G=2 |
| 5 | shard | take `self.state` lock      | |
| 6 | shard | `last_seen != Some(2)`, replay against S1 | backings registered against S1's engines |
| 7 | shard | set `last_seen = Some(2)`   | recorded as "fully registered for gen 2" |
| 8 | shard | return S1                   | this call's I/O uses S1, which is fine for this call |

The acquire load at (B) does synchronize with the writer's `fetch_add` at (W2), so anything the writer did before (W2) is visible to the reader after the acquire. But the reader's load at (A) happens before the acquire in program order, and the acquire cannot make (A) see S2 after the fact. `(S1, G=2)` is a legal observation.

After step 8, the cache says "fully registered for gen 2", but the actual registrations were only applied to S1. The next call from any shard thread:

| step | thread | action | result |
|------|--------|--------|--------|
| 9  | shard | (A) `current.load_full()` | sees S2 |
| 10 | shard | (B) `generation.load`     | sees G=2 |
| 11 | shard | `last_seen == Some(2)`    | no replay |
| 12 | shard | return S2                 | caller does I/O through S2 |

S2 routes through engines that this shard's `Backing` was never registered against. If S2 added a new disk, the new engine has no record of this shard's `base/len`. Once the data path actually issues I/O, the underlying `UringBlockDevice` looks up the buffer in its list of registered slots, does not find it, and the I/O fails. The bug stays broken until the next `apply_engines` advances the generation again, which forces a replay.

### Why `register_backing` does not close the window

`register_backing` ([`shard_view.rs:81-97`](https://github.com/Azure/unbounded/blob/971f7ccd610a0aef8df6569c2d684a64705b1aa4/cmd/unbounded-storage/src/disk_supervisor/shard_view.rs#L81-L97)) sets `last_seen_generation = None` and re-enters `current_or_replay`. That handles the same-thread sequence "push a new entry, then make sure it gets registered against whatever is current". It does not protect against a concurrent `apply_engines` between (A) and (B) inside the re-entered call. The same race fires there; it just starts from `last_seen = None` instead of `last_seen = Some(old_gen)`.

### Why the old pointer comparison was correct

The previous code compared `Arc::as_ptr(&ls)` against a cached pointer. The marker was always derived from the same `Arc` that the replay ran against, so after a replay the cache said "I have registered against the snapshot I am currently holding". If a later `apply_engines` published S2 at a different address, the marker for S2 was guaranteed to mismatch on the next call.

There is no ABA risk in this scheme. `current_or_replay` holds the `Arc<LocalStorage>` for the duration of the comparison, so S1's address cannot be reused while the reader is looking at it.

The real fix the previous code needed was the lock-ordering one (fold `registered` and the marker into one mutex), which 84685f0 also did. Swapping the comparison from pointer to generation was the part that introduced this race.

### Fix options

**Option 1: bundle the generation into the published snapshot.** Make the published unit `(snapshot, generation)` so `current()` returns both atomically:

```rust
struct LocalStorageSnapshotInner<B: BlockDevice + 'static> {
    ls: LocalStorageSnapshot<B>,
    gen: u64,
}

// in apply_engines:
let gen_n = self.generation.fetch_add(1, Ordering::AcqRel) + 1;
self.current.store(Arc::new(LocalStorageSnapshotInner {
    ls: snapshot,
    gen: gen_n,
}));
```

The reader does one `load_full` and gets a consistent pair. No ordering reasoning needed. `generation()` can stay as an observability accessor that reads the same inner.

**Option 2: read the generation before the snapshot.** Smaller patch:

```rust
let gen_n = self.topology.generation();   // (B) first
let ls = self.topology.current()?;         // (A) second
```

The writer publishes snapshot then bumps generation. So if the reader sees the new generation at (B), the writer's `fetch_add` happens before the reader's acquire, which means the writer's earlier `store` is also visible; the reader's later snapshot load is guaranteed to see the new snapshot. If the reader sees the old generation at (B), it may see either the old or the new snapshot at (A); if it sees the new snapshot it over-replays and records the old generation, and the next call re-triggers. Either way there is no under-replay. This depends on the producer's "store before bump" ordering, which would need a one-line comment in `apply_engines` so a future edit does not reorder it.

**Option 3: revert to pointer comparison.** Keep `ReplayState` (the lock-ordering fix), but move the marker back to `Option<*const LocalStorage<B>>`. Drop the use of `generation()` from the staleness check; keep it as a logging accessor.

Option 1 is the cleanest. Option 2 is the smallest patch. Option 3 is the most conservative.

### How to write a test for this

The race is hard to hit with a noop scheduler because `apply_engines` runs to completion before anything else gets to poll. Two options:

1. Hand-rolled test with two OS threads and a `Barrier`. The writer thread loops `apply_engines(vec![(new_path, new_engine)])`; the reader thread loops `current_or_replay`. After N iterations, assert that for every engine reachable through the currently-published `ls`, `MockDevice::registered_count()` includes this shard's backing. The `replays_every_backing_against_newest_local_storage` test already uses these accessors.

2. DST under `tests/disk_supervisor/`. Per `cmd/unbounded-storage/AGENTS.md`, this is the project's pattern for concurrent behavior. Higher cost but follows the convention.

(1) is enough to demonstrate the bug. (2) is what `AGENTS.md` would ask for once `LiveDiskTopology` has any other interleaving concerns to cover.

### Why this matters now

Today the data path does not issue I/O through `LiveShardLocalStore::{read_page, write_page}`. The only path exercised is `register_pages` at shard bring-up, which runs before any reload-driven `apply_engines`. The race does not fire today. It becomes visible as soon as:

- Phase 6 wires `Pool` callers to actually pull through `LiveShardLocalStore`, and a config reload fires while traffic is in flight.
- A test starts exercising the read/write paths concurrently with reloads.

The reason to fix it now rather than later is that the comment in the code at [`shard_view.rs:105-113`](https://github.com/Azure/unbounded/blob/971f7ccd610a0aef8df6569c2d684a64705b1aa4/cmd/unbounded-storage/src/disk_supervisor/shard_view.rs#L105-L113) claims the check is correct. The next maintainer who reads that comment will likely build on it, and the bug will be one step further from the code that introduced it.

jveski:
> **Heads Up: Experiment in AI Code Review**

Fixed!

cchildress (dismissed):

cchildress (approved):

PR #187, opened 2026-05-27
agent: setup daemon controller authnz

bcho:
## Summary
- Add reusable daemon credential management that issues kubelet-shaped client certificates with a daemon-specific group.
- Add a reusable CSR approver and wire Machina to approve daemon bootstrap and renewal CSRs with Machine/Node binding checks.
- Wire the agent daemon to use rotated daemon-controller credentials and add expiration to generated bootstrap tokens.
- Keep the current e2e bootstrap-token setup in place for now.

## Validation
- go test ./pkg/agent/daemoncred
- go test ./cmd/machina/machina/controller ./pkg/agent/daemoncred ./cmd/agent/internal/daemon


cmd/machina/machina/controller/daemon_csr_approver.go
```diff
@@ -0,0 +1,131 @@
...
+	c client.Client,
+	kubeClient kubernetes.Interface,
+) (*daemoncred.CSRApproverReconciler, error) {
+	claims := &daemonCSRClaimChecker{Client: c}
+	approver, err := daemoncred.NewCSRApprover(daemoncred.CSRApproverOptions{
+		BootstrapGroup:     bootstrapGroup,
+		DaemonGroup:        daemonGroup,
+		AuthorizeBootstrap: claims.bootstrapTokenMayClaimNode,
+		AuthorizeRenewal:   claims.nodeHasMachineBinding,
+	})
```
plombardi89, line 44: This approver handles the shared signer `kubernetes.io/kube-apiserver-client`. In `daemoncred.CSRApprover.Evaluate`, if a CSR uses that signer but doesn't look exactly like a daemon CSR, it gets **denied** instead of ignored. The reconciler then writes a `Denied` condition on it.

That signer is used by lots of things. Any other CSR for it (admin user certs, third-party tools, future controllers) gets denied by Machina the moment it shows up. That's a quiet way to break unrelated workflows.

Two ways to fix it:

1. In `pkg/agent/daemoncred/approver.go`, ignore CSRs that don't match instead of denying them. Only deny when the CSR clearly belongs to this approver - for example, the requester is `system:bootstrap:*`, or the requester is exactly `system:node:<nodeName>` and the cert organizations include `DaemonGroup`.
2. Use a dedicated signer name and run a matching signer. That gives full isolation.

`TestCSRApproverStockKubeletCertCannotRenew` in `approver_test.go` documents the current behavior. This should be fixed before turning the approver on in any cluster that runs other CSR-emitting workloads.

cmd/machina/machina/controller/daemon_csr_approver.go
```diff
@@ -0,0 +1,131 @@
...
+	nodeName string,
+) (bool, error) {
+	tokenID := strings.TrimPrefix(csr.Spec.Username, daemoncred.BootstrapUserPrefix)
+	if tokenID == "" {
+		return false, nil
+	}
+
+	secretName := bootstrapSecretPref + tokenID
+	var token corev1.Secret
+	if err := c.Get(ctx, client.ObjectKey{Namespace: metav1.NamespaceSystem, Name: secretName}, &token); err != nil {
```
plombardi89, line 64: `mgr.GetClient()` returns a cached client. The first `Get` on a `*corev1.Secret` makes controller-runtime start watching **every** Secret in the cluster (the cache can't filter by namespace on `Get` unless it's configured to). That's a big jump in memory and CPU for Machina, and Machina now keeps copies of every Secret in the cluster.

Fix: configure the cache in `manager.go` to only watch the namespaces Machina actually uses. Today that's `kube-system` (this PR's bootstrap-token lookup, plus `MachineReconciler.getBootstrapToken` at `machine_controller.go:472`) and `unbounded-kube` (`getSecretValue` at `machine_controller.go:451`):

```go
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
    Scheme: scheme,
    Cache: cache.Options{
        ByObject: map[client.Object]cache.ByObject{
            &corev1.Secret{}: {
                Namespaces: map[string]cache.Config{
                    metav1.NamespaceSystem:       {},
                    SecretNamespaceUnboundedKube: {},
                },
            },
        },
    },
    ...
})
```

This is also an improvement over what's on `main`, since the SSH-key path already had the same cluster-wide caching footprint.

cmd/machina/machina/controller/daemon_csr_approver.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	siteName := strings.TrimSpace(token.Labels[unboundedv1alpha3.MachineSiteLabelKey])
+	isDefaultToken := strings.EqualFold(strings.TrimSpace(token.Labels[defaultBootstrapTokenLabel]), "true")
+	if siteName == "" && !isDefaultToken {
+		return false, nil
+	}
+
+	var machines unboundedv1alpha3.MachineList
+	if err := c.List(ctx, &machines); err != nil {
```
plombardi89: Both `bootstrapTokenMayClaimNode` and `nodeHasMachineBinding` call `c.List(&MachineList{})` on every CSR event. That walks every Machine in the cluster. If the denial issue in the `NewCSRApprover` thread isn't fixed, every CSR for the shared signer also kicks off a reconcile, so the cost adds up fast.

Fix: register field indexes via `mgr.GetFieldIndexer().IndexField()` on `spec.kubernetes.bootstrapTokenRef.name` and `spec.kubernetes.nodeRef.name`. Then look up Machines with `client.MatchingFields{...}` instead of listing them all.

cmd/machina/machina/controller/daemon_csr_approver.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		// During early bootstrap, the Machine may not exist yet. When a Machine
+		// already references this token, enforce token-to-Machine-to-Node binding.
+		if siteName != "" && machine.Labels[unboundedv1alpha3.MachineSiteLabelKey] != siteName {
+			return false, nil
+		}
+
+		return machineNodeName(&machine) == nodeName, nil
+	}
```
plombardi89: The loop here stops at the first Machine that references this token. If that Machine's site or node doesn't match, it returns `false` right away without checking any other Machines. List order isn't stable, so the answer can change between reconciles.

Concrete bug: Machine A references token X with `NodeRef=node-a`, Machine B references token X with `NodeRef=node-b`. A real CSR from `node-b` gets denied because A happens to be listed first.

Two options: (a) keep looking through the rest of the Machines and approve if any of them matches this node, or (b) enforce that a token may only be referenced by one Machine elsewhere in the code.

cmd/machina/machina/controller/daemon_csr_approver.go
```diff
@@ -0,0 +1,131 @@
...
+			continue
+		}
+
+		// During early bootstrap, the Machine may not exist yet. When a Machine
+		// already references this token, enforce token-to-Machine-to-Node binding.
+		if siteName != "" && machine.Labels[unboundedv1alpha3.MachineSiteLabelKey] != siteName {
+			return false, nil
+		}
+
+		return machineNodeName(&machine) == nodeName, nil
```
plombardi89: `machineNodeName` falls back to `machine.Name` when `NodeRef` is nil. The "Machine name equals Node name" rule (see `machine_controller.go:798`) only holds after Machina has populated `NodeRef`. On first bootstrap, `NodeRef` is nil, and `buildMachineCR` in `cmd/agent/internal/daemon/daemon.go` names the Machine after `cfg.MachineName`, not `cfg.NodeName`.

If those two values differ - or the operator pre-creates a Machine with some other name - this CSR gets denied with no clear error.

Options: (a) document and enforce that `Machine.metadata.name == agent.NodeName` until `NodeRef` is filled in; or (b) relax bootstrap so the site-scoped token check is enough until `NodeRef` shows up.

Related: in renewal, `nodeHasMachineBinding` does `Get(name=nodeName)` first and hits the same fallback. An operator who clears `NodeRef` thinking they've cut off this daemon's renewals hasn't actually done that. Worth a doc note.

internal/kube/bootstrap_token.go
```diff
@@ -18,17 +19,19 @@ import (
 )
 
 const (
-	bootstrapTokenAlphabet = "abcdefghijklmnopqrstuvwxyz0123456789"
+	bootstrapTokenAlphabet   = "abcdefghijklmnopqrstuvwxyz0123456789"
+	DefaultBootstrapTokenTTL = 3 * time.Hour
```
plombardi89: Bootstrap tokens used to never expire. Now they expire in 3 hours. The only non-test caller is `ensureBootstrapToken` in `cmd/kubectl-unbounded/app/site_init.go`. It reuses any non-expired token and only creates a new one once **all** tokens have expired. Nothing rotates them ahead of time.

Real failure: operator runs site init, gets a token, ships a bare-metal node. The node powers on 4 hours later. The token is already gone, so the daemon can't bootstrap. 3 hours is also tight for `e2e.py` if anything slows down.

Pick one: (a) add a rotation loop in `site_init`, (b) rotate when remaining lifetime drops below a threshold, or (c) raise the default to 24 hours (kubeadm's default) and document the expectation.

cmd/agent/internal/daemon/daemon.go
```diff
@@ -73,16 +115,48 @@ func run(ctx context.Context, log *slog.Logger, newClient kubeClientFunc, nodeOp
...
+	}
+
+	opts := daemoncred.ControllerCertificateOptions{
+		Name:          daemonControllerCertificateName,
+		SignerName:    daemoncred.DefaultControllerCertificateSignerName,
+		DaemonGroup:   daemonControllerGroup,
+		CredentialDir: runOpts.DaemonCredentialDir,
+	}
+
+	provider, err := daemoncred.NewRESTConfigProvider(ctx, bootstrapCfg, agentCfg.NodeName, opts)
```
plombardi89, line 144: `daemoncred.NewRESTConfigProvider` is called here with default options. That gives the daemon only 10 seconds (`defaultControllerCertificateWaitTimeout` in `pkg/agent/daemoncred/credentials.go:33`) to get its certificate, with no retry. In those 10 seconds: the CSR has to be created, Machina has to see and approve it, kube-controller-manager has to sign it, and client-go's cert manager has to pick it up. A short Machina leader-election blip or any apiserver lag is enough to fail daemon startup.

Fix: either set `WaitTimeout` on the options literal here (a value in the 60-120 second range covers typical apiserver and controller hiccups), or wrap `daemonControllerCredentials` in a retry loop.

cmd/machina/machina/controller/daemon_csr_approver.go
bcho: updated to deny only to csr asking for node group + `<daemonGroup>` group, ignore other unknown csr from the signer

cmd/machina/machina/controller/daemon_csr_approver.go
bcho: updated to add node ref index

cmd/machina/machina/controller/daemon_csr_approver.go
bcho: remove the fallback logic, now in the renewal case, we require the node is bound to a machine via node ref

internal/kube/bootstrap_token.go
bcho: increased to 24hr

cmd/agent/internal/daemon/daemon.go
bcho: updated timeput settings

cmd/machina/machina/controller/daemon_csr_approver.go
bcho: updated the logic and clarify the behavior: the bootstrap token is shared across machine within the site. As long as the requesting bootstrap token is matching with the site and within expiry time, we will allow the usage.

cmd/machina/machina/controller/daemon_csr_approver.go
bcho: added cache filter for this

plombardi89 (approved):

PR #210, opened 2026-06-03
Add S3 compatible frontend/backend to unbounded-storage

jveski:
Allows clients to access S3 objects from Orca through the unbounded-storage p2p cache mesh. Rebases/expands #195


plombardi89 (commented):
## Review: Add an S3-compatible frontend/backend to unbounded-storage

Overall this is a careful, well-documented change. It lets clients read objects using the S3 protocol on top of the existing storage system. The structure is sound, and the tests covering the helper functions, XML/header formatting, and the smoke test (which exercises both protocols) are thorough. The items below are worth addressing before merge. None of them undermine the core design.

### High - Unused crypto dependencies (`sha2`, `hmac`)

`Cargo.toml`/`Cargo.lock` add `sha2` and `hmac` (and their transitive dependencies), but nothing in the source uses them - a search for `sha2`, `hmac`, `sigv4`, and `authorization` returns no matches. The S3 backend deliberately sends unsigned requests, and a test even asserts that no `authorization` or `x-amz-date` headers are present (s3.rs:1988). These dependencies should be removed until request signing is actually implemented. Adding crypto libraries to the build and supply-chain surface for unused code is not justified.

### High - `OriginRing::WorkerLocal` assumes one memory region per thread

In `origin_ring.rs:1067-1091`, the thread-local `WORKER_RING` builds the ring and registers its memory region on first use, then caches it for the life of the thread, ignoring the `region` passed on later calls. This is safe under the current model, where each inbound request is handled by a freshly spawned, single-use thread (`fabric/rpc.rs:244-251`). But it carries two problems:

- **Latent correctness risk.** Each storage shard has its own separate scratch memory. If the request-handling threads are ever pooled or reused across shards, a reused thread would keep the first shard's registered region while a later request computes its write offsets against a different shard's memory - sending received bytes into the wrong place. The comment "a reused worker thread keeps it for subsequent requests" invites exactly that scenario. At minimum, add an assertion that a reused ring's region matches the one requested, or key the cache by region.
- **The optimization currently does nothing.** Because threads are not reused today, the cache never hits, so every cache-miss request pays the full cost of setting up and tearing down a ring (including buffer registration). Worth a comment noting it is forward-looking, or reconsidering it.

### Medium - Substantial duplication between `s3.rs` and `http.rs`

`s3.rs` (1056 lines) re-implements, almost verbatim, logic that already exists in `http.rs`: `pages_capacity`, `page_byte_offset`, `locate_in_pages`, `write_slice_into_pages`, `zero_fill_pages_from`, `copy_body_into_pages`, `recv_chunk`, `io_to_err`, `clone_sockaddr`, `absolute_range`, `expected_body_len`, `TcpConn`, and the whole fetch/length/stream skeleton - along with duplicated copies of their tests. This includes duplicated low-level pointer arithmetic, which is the most dangerous kind of code to keep in sync: a fix applied to one copy can easily be missed in the other. The PR already establishes the right pattern (the shared `http/server.rs` and `origin_ring.rs`); the same should be done here by moving the shared fetch helpers and `TcpConn`/IO helpers into a common file. The real differences between the S3 and HTTP paths are small (a `Host` header sourced from a hostname, and mapping 404 to `OriginNotFound`), so the two backends could share nearly all of this.

### Medium - Removing TLS config is a coupled, potentially breaking change

The change deletes `TlsCfg` and the `FrontendSpec.tls` field. `BackendSpec` uses `deny_unknown_fields` (the `unknown_fields_rejected_in_backend` test depends on it); if `FrontendSpec` does too, then any existing config that still includes a `[frontends.tls]` block - previously accepted and ignored - will now fail to load. That is a behavioral regression bundled into an S3 feature PR. Either continue accepting and ignoring the `tls` block, or call this out explicitly as an intentional breaking change in the PR description. It also seems unrelated to S3 and may be better as its own separate change.

### Low

- **`cancel_and_drain` can return while the operation is still live (core.rs:4512-4545).** On reaching `MAX_DRAIN_ITERS`, it breaks out with only a `debug_assert!`. In release builds this means it can return while the kernel may still be writing into the destination page - the exact use-after-free the function exists to prevent. Since each iteration blocks on a real completion, it would be safer to keep draining (or abort) rather than return; the bound should be a safety net, not an escape hatch.
- **The `bucket` field is unused.** It is accepted in config but has no effect ("reserved, not yet wired"). Config that does nothing can mislead operators; consider adding it only once it takes effect.
- **`absolute_range` uses saturating arithmetic (s3.rs:1851).** A bad `stripe_idx`/`stripe_size` would silently produce a wrong range rather than an error. The geometry is internal so this is defensive, but a checked variant with an explicit error would be safer.
- **No deterministic-simulation (DST) coverage for the new concurrent paths.** The inline unit tests and the `network.rs` test `drop_in_flight_recv_fixed_drains_the_op` are good, but the crate names DST as its primary correctness gate for concurrent code. Acceptable if the existing HTTP backend also lacks it, but worth noting.
- **`S3Backend::backend_id()` appears unused** in this change; confirm it is needed or drop it.

### Questions for the author
1. Is unsigned/public-bucket access the intended scope for now, with request signing (and the `sha2`/`hmac` dependencies and `bucket` field) deferred? If so, please remove the unused dependencies until the signing code lands.
2. Is the removal of `FrontendSpec.tls` an intentional breaking change? Does any deployed config still set it?
3. Can the request-handling threads ever be pooled or reused across shards in the future? That determines whether the `WorkerLocal` region caching is a real hazard or only a latent one.

### Summary
The S3 path itself looks correct and is well tested, so nothing here blocks the core design. The most important items are removing the unused dependencies, guarding the `WorkerLocal` region assumption, reducing the duplicated fetch logic, and clarifying the TLS config removal.

plombardi89 (approved):
Discussed feedback and plan is to refactor and fix in follow ups to keep work well isolated

PR #224, opened 2026-06-08
Preserve route metadata and document Cilium migration

chokevin:
## Why

During the Azure Linux/CNI investigation, the route manager was found to drop route metadata that callers explicitly set. Separately, the live incident ultimately traced to stale Cilium state during CNI takeover, so the normal migration guidance should point operators at Cilium's own cleanup flow rather than relying on ad hoc repair.

The same live validation also showed low-rate WireGuard `rx_errors` can persist while peers, handshakes, Ray smoke, and pod datapath are healthy. Those receive-only counter deltas should stay visible in logs/diagnostics but should not force node status to `Errors` when the corresponding peers are healthy.

## What changed

- Preserve `Flags` and `ScopeGlobal` while `UnifiedRouteManager.SyncRoutes` normalizes and merges desired routes.
- Track those fields in installed route state so drift detection catches flag/scope changes.
- Add regression coverage for gatewayless `unbounded0` route metadata preservation and route update detection.
- Document Cilium-to-unbounded migration steps, including `cilium-dbg post-uninstall-cleanup --all-state -f` and checks for stale `cilium_host`, fwmark rules, `cil_sock*`, and `cilium_lb*` state.
- Suppress WireGuard `rx_errors`-only link-stat node errors when all peers on that interface are healthy, while preserving tx/drop and unhealthy-peer link-stat warnings as node errors.

## Notes for reviewers

The Cilium cleanup is intentionally documentation only in this PR. The live repair proved stale Cilium host/service-LB state can break ClusterIP DNS after takeover, but full Cilium cleanup should be handled by Cilium tooling or a separately scoped operational repair.

The link-stats change is intentionally narrow: it only suppresses `rx_errors`-only warnings for interfaces whose peers are healthy. `tx_errors`, drops, missing peers, or unhealthy peers still surface as node errors.

## Validation

- `go test ./internal/net/netlink -run 'TestRouteNeedsUpdate_FlagsChanged|TestRouteNeedsUpdate_ScopeGlobalChanged|TestBuildInstalledState|TestBuildKernelRoute_GatewaylessScopeGlobal|TestSyncRoutes_PreservesGatewaylessRouteMetadata' -count=1 -v` in a Linux container
- `go test ./cmd/unbounded-net-node -run 'TestLinkStatsWarningsSuppressHealthyWireGuardRxErrors|TestLinkStatsWarningsKeepWireGuardRxErrorsWhenPeerUnhealthy|TestLinkStatsWarningsKeepWireGuardTxAndDropErrors|TestLinkStatsWarningsKeepRxErrorsWithNoMatchingPeer' -count=1 -v` in a Linux container
- `GOOS=linux GOARCH=amd64 golangci-lint run -c .golangci.yaml -E wsl_v5 ./cmd/unbounded-net-node ./cmd/unbounded-net-controller ./cmd/kubectl-unbounded`
- Linux-target compile checks for `cmd/unbounded-net-node`, `cmd/unbounded-net-controller`, and `cmd/kubectl-unbounded`
- Earlier validation also covered Linux-target compile checks for `internal/net/netlink`, `internal/net/ebpf`, and `cmd/unroute`

AI assistance was used to prepare this change; the final diff and validation were reviewed before opening the PR.

### DIFF
--- a/cmd/unbounded-net-node/status_server.go
+++ b/cmd/unbounded-net-node/status_server.go
@@ -2199,13 +2199,6 @@ func (s *nodeStatusServer) getNodeStatus() *NodeStatusResponse {
 		NodeErrors: filterExpiredNodeErrors(s.state.nodeErrors, time.Now(), time.Minute),
 	}
 
-	// Append link stats warnings as node errors if any interfaces have incrementing counters.
-	if s.state.linkStatsMonitor != nil {
-		for _, w := range s.state.linkStatsMonitor.GetWarnings() {
-			status.NodeErrors = append(status.NodeErrors, NodeError{Type: "link-stats", Message: w})
-		}
-	}
-
 	// Append kube-proxy health warning if the health check is failing.
 	if s.state.kubeProxyMonitor != nil {
 		if w := s.state.kubeProxyMonitor.GetWarning(); w != "" {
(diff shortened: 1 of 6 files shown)


chokevin:
Superseded by smaller split PRs: route metadata, Cilium migration docs, and WireGuard link-stat status semantics.

PR #225, opened 2026-06-08
Fix an issue where fabric tests could bind the wrong interface and fail.

cchildress:
The tests that depended on libfabric were failing for me, but not always others. Turns out my test runs were binding an extra interface I have in WSL. We should have the tests always binding the loopback.

### DIFF
--- a/cmd/unbounded-storage/src/fabric/fabric.rs
+++ b/cmd/unbounded-storage/src/fabric/fabric.rs
@@ -460,6 +460,11 @@ mod tests {
         // the test lean.
         cfg.progress_threads = 1;
         cfg.max_inflight = 16;
+        // Pin the source bind to loopback so self_address() is
+        // deterministic on multi-NIC hosts (see tcp_loopback_cfg in
+        // fabric/tests.rs for the full rationale).
+        cfg.listen = true;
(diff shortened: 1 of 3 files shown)


cmd/unbounded-storage/src/fabric/fabric.rs
```diff
@@ -460,6 +460,11 @@ mod tests {
         // the test lean.
         cfg.progress_threads = 1;
         cfg.max_inflight = 16;
+        // Pin the source bind to loopback so self_address() is
+        // deterministic on multi-NIC hosts (see tcp_loopback_cfg in
+        // fabric/tests.rs for the full rationale).
+        cfg.listen = true;
+        cfg.listen_addr = Some("127.0.0.1:0".to_string());
```
jveski, line 467: Should this be `:0` to support scenarios where the tcp transport is used outside of tests? e.g. on nodes that don't support rdma

cchildress: This is in a test block so it shouldn't be called outside of tests.

jveski: too Go brained :skull:

jveski (approved):

PR #227, opened 2026-06-08
Document Cilium to unbounded migration cleanup

chokevin:
## Why

The flex incident showed that stale Cilium state can keep owning pod CIDR and ClusterIP traffic after Cilium is no longer the intended CNI. The safest normal migration path is to use Cilium's cleanup tooling before unbounded-net takes over, rather than relying on ad hoc live BPF map repair.

## What changed

- Document a Cilium-to-unbounded migration flow using `cilium-dbg post-uninstall-cleanup --all-state -f`.
- Add checks for stale `cilium_host`, Cilium fwmark rules, `cil_sock*`, and `cilium_lb*` state.
- Document the symptoms observed when cleanup is skipped or incomplete.

## Validation

- `git diff --check origin/main...HEAD`

AI assistance was used to prepare this change; the final diff was reviewed before opening the PR.


bcho (dismissed):
IMO reimaging the host before switching network setup could be a better choice, but I think it's nice to have to include in this doc

chokevin:
> [Jun 8, 2026 ](#pullrequestreview-4452813082)
> [ View reviewed changes ](/Azure/unbounded/pull/227/files/7503f68d86828064914c4723b023edf42649e0d8)

Yeah I think thats fair.  Re-image would have probably solved a lot of this out of the box but I didn't. Can't always expect your customers to do the right thing all the time :)

cchildress (approved):

bcho (approved):

bcho (approved):

phealy (approved):

Issue #235, opened 2026-06-09
release-upgrade silently wipes machina-config apiServerEndpoint, crashing the controller


plombardi89:
## Summary

Every `release-upgrade.yaml` deploy silently overwrites the `machina-config` ConfigMap with the bundled `deploy/machina/03-config.yaml.tmpl`, which is missing the per-cluster `apiServerEndpoint` field that `machina-controller` v0.1.5+ requires. The controller then exits at startup with `error: resolve cluster info: API server endpoint not set in config` and stays in `CrashLoopBackOff` until an operator manually re-patches the ConfigMap.

The only code path that populates `apiServerEndpoint` is `kubectl unbounded site init`, which runs once at cluster bootstrap. Any subsequent upgrade through the release-upgrade workflow undoes it.

## How this surfaced

`unbounded-stable` had been stuck on `v0.1.5` for ~4 weeks because every release-upgrade run since v0.1.3 failed at the asset-download step (fixed in #233). When the first successful release-upgrade run after #233 deployed v0.1.11, the rollout failed at `Wait for rollouts`: the new `machina-controller` pod was in `CrashLoopBackOff` for the same reason as the existing v0.1.5 pod had been for the past month. Investigation showed the cluster has had a missing-config-field problem since 2026-05-13, hidden by the asset-download failure masking every later deploy.

A manual `kubectl apply` to re-add the field unbroke the cluster, but the underlying workflow bug means the next release-upgrade run (for any future tag) will silently break it again.

## Root cause

`cmd/machina/machina/controller/config.go`:

```go
type Config struct {
    APIServerEndpoint       string        `yaml:"apiServerEndpoint"`
    MetricsAddr             string        `yaml:"metricsAddr"`
    ProbeAddr               string        `yaml:"probeAddr"`
    EnableLeaderElection    bool          `yaml:"enableLeaderElection"`
    MaxConcurrentReconciles int           `yaml:"maxConcurrentReconciles"`
    ProvisioningTimeout     time.Duration `yaml:"provisioningTimeout"`
}
```

`cmd/machina/machina/controller/cluster_info.go:50` makes the field mandatory:

```go
if cfg.APIServerEndpoint == "" {
    return nil, fmt.Errorf("API server endpoint not set in config")
}
```

`deploy/machina/03-config.yaml.tmpl` does NOT include the field:

```yaml
data:
  config.yaml: |
    metricsAddr: ":8080"
    probeAddr: ":8081"
    enableLeaderElection: false
[...]

PR #236, opened 2026-06-09
release-upgrade: workaround #235 by merging bundled machina-config with live

plombardi89:
## Why

Temporary workaround for #235.

The release tarball ships a `machina-config` ConfigMap whose inner `config.yaml` is missing the per-cluster `apiServerEndpoint` field that `machina-controller` v0.1.5+ requires. Only `kubectl unbounded site init` populates that field on the live cluster (and explicitly skips the bundled template via `skipPaths`). The release-upgrade workflow does not do that skip - it runs `kubectl apply --server-side --force-conflicts -R -f .../machina/` which actively overwrites the live ConfigMap with the empty-of-apiServerEndpoint bundled template, crashing the controller on the next pod restart.

That is what happened to `unbounded-stable` on 2026-05-13 and kept the cluster broken for 26 days (every subsequent release-upgrade run failed at the asset-download step (#233), masking the underlying controller crash). The v0.1.11 backfill after #233 surfaced the crash and required a manual ConfigMap patch to unblock.

## What changes

Add a single step to `release-upgrade.yaml`, immediately before the existing `Apply manifests (upgrade)` step:

```
Workaround for #235 - merge bundled machina-config with live ConfigMap
```

It:

1. Reads the live `machina-config` ConfigMap (`--ignore-not-found` for fresh-cluster bootstrap).
2. Parses both the bundled `data.config.yaml` and the live `data.config.yaml`.
3. Merges with **policy A**: live values win on conflicts; bundle contributes new keys only.
4. Rewrites the bundle file in place so the existing `Apply manifests (upgrade)` step pushes the merged manifest.
5. **Strict** YAML round-trip verification at the end; any parse failure aborts the workflow.

If the bundled template ever stops shipping `03-config.yaml` (e.g. when issue #235 is fixed via option (ii)), the step emits a `::warning::` and exits cleanly so the workflow does not break.

## Tested locally against three fixtures

```
=== Test: live has overrides; bundle adds nothing new ===
Preserved 5 key(s) from live ConfigMap
No new keys from bundle
[live's apiServerEndpoint and overridden maxConcurrentReconciles preserved]

=== Test: bundle introduces a new key ===
Preserved 5 key(s) from live ConfigMap
Added 1 new key(s) from bundle: reachabilityProbeInterval
[live values preserved; new key appended]

=== Test: no live ConfigMap (fresh cluster) ===
No live machina-config ConfigMap; using bundle as-is
[bundle applied unchanged]
```

The merged manifest uses block-scalar (`|`) style for the inner `config.yaml` so it stays readable in the run log.

## Caveats (documented in the step's header comment)

- **Comments in the bundled template are lost.** PyYAML's `safe_dump` doesn't preserve them. The active config values are what matter for controller behavior.
- **Does not patch a cluster whose `apiServerEndpoint` has already been wiped.** If the field is missing when this step runs (e.g. left over from before this PR), the merge produces a manifest without the field, the apply pushes it, and the controller stays broken. `unbounded-stable` was hand-patched on 2026-06-08 to add the field back; this workaround prevents future runs from wiping it again.
- **Behavior on bundle changing the default of an existing key.** Policy A means the live value wins. If a future release changes `maxConcurrentReconciles` default from 50 to 100, an existing cluster keeps its current value. Operators wanting the new default must update the ConfigMap manually. Acceptable tradeoff for a workaround.

## Validation

- `actionlint .github/workflows/release-upgrade.yaml` clean.
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release-upgrade.yaml'))"` parses; 4 jobs intact.
- Fixture-based local testing of the Python merge logic across three scenarios (above).

## What this PR is NOT

This is a temporary workaround scoped to our deploy workflow, not a fix to the underlying bug. The real fix is tracked in #235 and would land in one of:

- The controller falling back to in-cluster API discovery so `apiServerEndpoint` becomes optional.
- The release pipeline dropping `03-config.yaml` from the tarball.
- A `kubectl unbounded site upgrade` subcommand that owns the skip semantics.

When any of those land, **remove this step entirely** (the comment block above the step says so explicitly).

## Related

- #233 - fixed the upstream workflow trigger so deploys actually attempt; this surfaced the underlying controller crash.
- #234 - fixed the smoke-discover step that was checking out an older tag.
- #235 - the underlying bug being worked around.

### DIFF
--- a/.github/workflows/release-upgrade.yaml
+++ b/.github/workflows/release-upgrade.yaml
@@ -345,6 +345,125 @@ jobs:
             exit 1
           fi
 
+      - name: Workaround for #235 - merge bundled machina-config with live ConfigMap
+        if: env.MODE == 'upgrade'
+        # WORKAROUND for https://github.com/Azure/unbounded/issues/235.
+        #
+        # The shipped 03-config.yaml ships a machina-config ConfigMap whose
+        # inner config.yaml lacks the per-cluster `apiServerEndpoint` field
+        # required by machina-controller v0.1.5+. Only
+        # `kubectl unbounded site init` ever populates that field on the
+        # live ConfigMap. If we let the bundled template apply unmodified,
+        # it silently overwrites the live value and the controller starts
+        # crashing on the next pod restart.
+        #
+        # Instead of skipping the file outright (which would also drop any
+        # new keys the bundle introduces), we MERGE: load the live
+        # ConfigMap's inner config.yaml, load the bundle's inner
+        # config.yaml, take the union with LIVE WINNING on conflicts
+        # (policy A), and rewrite the bundle file in place. The subsequent
+        # `kubectl apply` then pushes a manifest that preserves all
+        # operator state while adding any new keys this release
+        # introduces. Strict YAML round-trip verification at the end
+        # aborts if the merged manifest doesn't reparse.
+        #
+        # Caveats:
+        # - YAML comments in the bundled template are lost (PyYAML
+        #   safe_dump does not preserve them). The active values are
+        #   what matter for controller behavior.
+        # - When this workaround is no longer needed, the step warns and
+        #   continues (apply proceeds normally).
+        #
+        # REMOVE this step (and this comment) once issue #235 lands
+        # upstream (e.g. controller falls back to in-cluster discovery, or
+        # the release tarball stops shipping 03-config.yaml).
+        env:
+          BUNDLE_FILE: ${{ env.MANIFESTS_DIR }}/machina/03-config.yaml
+        run: |
+          set -euo pipefail
+          if [[ ! -f "$BUNDLE_FILE" ]]; then
+            echo "::warning::Workaround for #235 expected ${BUNDLE_FILE} but it was not present; the workaround may no longer be needed"
+            exit 0
+          fi
+
+          # Fetch the live inner config (empty if the ConfigMap does not exist yet).
+          LIVE_INNER=$(kubectl -n unbounded-kube get cm machina-config \
+            -o jsonpath='{.data.config\.yaml}' --ignore-not-found)
+          export LIVE_INNER
+
+          python3 - <<'PY'
+          import os
+          import sys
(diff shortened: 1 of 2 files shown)


plombardi89:
Extracted the inline Python heredoc into a standalone `hack/release/merge-machina-config.py` script (commit 3cb743b).

Net changes since the last review pass:

1. **New file `hack/release/merge-machina-config.py`** - mode 0755, shebang, module docstring covering the contract, the policy A semantics, and the comment-loss caveat. Reads `BUNDLE_FILE` and `LIVE_INNER` from the environment, exits non-zero on any parse failure or round-trip failure.

2. **New step `Fetch release tooling from default branch (workaround for #235)`** - sparse `actions/checkout` of `hack/release/` from the default branch into `_release-tools/`. Necessary because the deploy job's primary checkout is pinned to the deployed tag; backfills of tags older than this PR (e.g. v0.1.11) would not have the script in the workspace otherwise. Mirrors the pattern from #234 for smoke scripts.

3. **Existing merge step shrunk** - the ~80-line inline heredoc collapses to a one-line `python3 "$MERGE_SCRIPT"` invocation. Header comment trimmed; the script-side docstring carries the detailed caveats now.

The workflow file went from +119 lines to +29 net. The script is +123 lines and can also be run locally for debugging:

```
BUNDLE_FILE=/path/to/03-config.yaml \
LIVE_INNER="$(kubectl -n unbounded-kube get cm machina-config -o jsonpath='{.data.config\.yaml}')" \
python3 hack/release/merge-machina-config.py
```

Locally re-validated:
- `actionlint .github/workflows/release-upgrade.yaml` clean
- Workflow YAML parses; 4 jobs intact
- Script syntax check passes
- Four fixture tests: live overrides + new bundle key (preserves overrides, picks up new key), fresh cluster (uses bundle as-is), missing `BUNDLE_FILE` (exits 1), invalid YAML (exits 1)

bcho (approved):

PR #247, opened 2026-06-10
Allow manual-bootstrap machine name to resolve at runtime

plombardi89:
## Summary

`kubectl unbounded machine manual-bootstrap` required a machine name as a positional argument, which was baked into the agent JSON config and rendered into the bootstrap script / cloud-init. That made the payload awkward to reuse with cloud-init across many instances (e.g. Azure VMSS or AWS Auto Scaling), because every instance shared one hard-coded name.

This change makes the machine name optional and resolvable on the host at startup.

## What changed

- **Optional name in the CLI** (`cmd/kubectl-unbounded/app/machine_manual_bootstrap.go`): `Args` is now `MaximumNArgs(1)`; `NAME` is validated as a DNS-1123 subdomain when supplied. The payload comment header shows `(resolved at runtime)` when omitted. Help text and a VMSS/ASG example added.
- **`MachineName` is now `omitempty`** (`pkg/agent/config/config.go`): when no name is supplied the field is dropped from the generated JSON entirely.
- **New `AgentConfig.BackfillMachineName`** returns the resolution source (`config` / `env` / `hostname`). Order: existing valid value -> `AGENT_MACHINE_NAME` -> host hostname. Env/hostname values are trimmed, lowercased, and validated (via a unit-testable `normalizeDerivedMachineName` helper). The value becomes the Machine CR name, so DNS-1123 validation is used.
- **Guaranteed resolution before any reader**: `BackfillMachineName` is called before `BackfillNodeName` (which falls back to `MachineName`) at every `AgentConfig` load site:
  - `normalizeConfig` (config load) - logger threaded through `loadConfig`/`normalizeConfig`; logs once only when the name is derived.
  - daemon `FindActiveMachine`.
  - defensively at the top of `goalstates.ResolveMachine`.
- **Legacy env path relaxed**: `MACHINA_MACHINE_NAME` is no longer required in `loadConfigFromEnv`; it falls through to the same backfill.
- **Docs**: new "Reusable bootstrap payloads (VMSS / Auto Scaling)" section in `docs/content/guides/agent.md`; `MachineName` marked optional in the agent configuration reference.

## Behavior for existing flows

- Supplying `NAME` (and the machina / metalman controller paths, which always set a name) is unchanged: source is `config`, no extra logging, payload byte-compatible apart from the now-optional field.
- The shared `unbounded-agent-install.sh` is untouched; resolution happens in the agent (Go), so there is no new runtime dependency (no `jq`/`sed`).

## Testing

- `make fmt`, `make lint` (0 issues), `CI=1 make test` (full suite passes).
- New/updated tests: `normalizeDerivedMachineName`, `BackfillMachineName` (config/env/hostname/invalid + ordering), `ResolveMachine` backfills `KubeMachineName`, `loadConfig` backfills when missing + env path optional + still requires other vars, manual-bootstrap empty-name validate/render/execute and invalid-name validation, and the `omitempty` JSON assertion.


plombardi89:
Addressed the review feedback about not changing `pkg/agent` (shared library with external consumers).

Follow-up commit `0aaf41c` moves all machine-name resolution out of `pkg/agent` and into `internal/provision` (the unbounded-specific config layer that already wraps `pkg/agent/config` and is imported by both the agent `start` path and the daemon):

- **`pkg/agent` is now byte-for-byte unchanged vs `main`** (`git diff origin/main -- pkg/agent/` is empty): no `omitempty` tag change, no `BackfillMachineName`, and the defensive call in `goalstates.ResolveMachine` is gone. External consumers of the library see the original behavior.
- New `internal/provision/ResolveMachineName(cfg) (source, err)` holds the logic (config -> `AGENT_MACHINE_NAME` -> hostname, lowercased + DNS-1123 validated).
- `cmd/agent` calls it before `BackfillNodeName` at both `AgentConfig` entry points (`normalizeConfig` and the daemon's `FindActiveMachine`), preserving the guarantee that `MachineName` is populated before any reader.
- Per the decision to keep `pkg/agent` untouched, the generated config now carries an empty `"MachineName"` (resolved on the host) instead of omitting the key.

**Why `internal/provision` and not a `cmd/agent`-local package:** the two `AgentConfig` entry points live in different packages - `normalizeConfig` in `cmd/agent/internal/cmd` and `FindActiveMachine` in `cmd/agent/internal/daemon` - so the resolver needs a shared home both can import. `cmd/agent/internal/cmd` already imports `cmd/agent/internal/daemon` (`start.go` uses `daemon.PersistAppliedConfig`, etc.), so putting the resolver in `daemon` would be the only cycle-free `cmd/agent`-local option, but that inverts the layering (config normalization reaching into the daemon package). `internal/provision` is the natural neutral spot: both packages already import it, it has no import cycle (`provision -> pkg/agent/config`, one way), it is internal to this module (no external consumers, unlike `pkg/agent`), and it already owns the unbounded-specific config layer (`UnboundedAgentConfig`, `BuildAgentConfig`).

`make fmt`, `make lint` (0 issues), and `CI=1 make test` (full suite) pass.

bcho (approved):

PR #267, opened 2026-06-16
unbounded-storage: stabilize config schema

jveski:
Reworks the unbounded-storage config schema to prepare for a stable release. The main change is modeling the request path as a series of "components" that source/sink each other.

The idea is that we can add new component types in the future to support more complex routing, rbac, etc.

It also adds support for multiple p2p "neighborhoods" in case we ever need to dedicate some subset of HCAs/NICs/disks to a specific use.


plombardi89:
## Code review

Reviewed the full diff (76 files, +7359/-6124). The component-graph rework (`backend -> cache -> neighborhood -> frontend` with oneof variants, multi-cache/neighborhood support, fake backend, loadgen frontend) is well-structured and thoroughly tested. Issues below, ordered by severity.

### Bugs

**1. (High, for the new feature) Multi-cache disks oversubscribe the same storage cores.** `DiskRegistrySet::reconcile_cache` hands every per-cache `DiskRegistry` a full clone of the shared `disk_slots`:

```rust
.or_insert_with(|| DiskRegistry::new(self.target.clone(), self.disk_slots.clone()))
```
(`cmd/unbounded-storage/src/storage/disks/mod.rs`)

Each registry runs `assign_disk_cpus` independently with no shared "used" tracking. With two caches that each own a disk, both pick `slots[0]` and pin two disks to the **same** storage core, leaving the other idle. This breaks the disjoint-CPU invariant the single `DiskRegistry` was designed to preserve (ARCHITECTURE.md 7.10). The PR's stated motivation is "dedicate some subset of HCAs/NICs/disks to a specific use," so multi-cache-with-disks is the intended path. Single-cache (the common case) is unaffected. The CPU slot pool needs to be assigned globally across caches, not per-cache.

**2. (High) `gen-storage-mesh-config.sh` emits an invalid `[startup.fabric]` section.** The generator renamed `listen_addr` -> `addr` but left it directly under `[startup.fabric]`:

```
[startup.fabric]
addr = "$LOCAL_IP:$OPT_PORT"
```

`FabricCfg` no longer has an `addr` field (the TCP bind moved into the `binds` oneof), so the daemon's strict TOML loader (`deny_unknown_fields`) rejects this. The smoke test, the inttest, and the Go supervisor all correctly use `[startup.fabric.binds.tcp]` / `fabric.tcp`; only this script was missed, so every config it generates fails to load.

### Architectural concerns

**3. (Medium) Multi-neighborhood fabric self-identity is inconsistent.** Peer ids are scoped per-neighborhood via `scoped_peer_id(neighborhood_id, node_id)`, but the node advertises a single fabric identity derived from only the alphabetically-first neighborhood (`local_self_peer` in `shard_layer.rs`). `self_peer` is sent as CM private data on every dial so the accepting peer keys the inbound connection. A node in neighborhoods A and B advertises `scoped(A, id)`, but peers in B expect `scoped(B, id)`, so cross-neighborhood keying won't match. Latent (multi-neighborhood is forward-looking), but the scoped-id design and the single self-id are at odds. Consider asserting single-neighborhood for now, or making self-identity per-endpoint.

**4. (Medium) Supervisor silently drops discovered peers when the ConfigMap declares no neighborhood.** `applyRing` iterates `cfg.Neighborhoods` and injects `local_node_id` + merged peers into each. If zero neighborhoods are declared, the discovered ring peers are dropped with no warning (`internal/storagesupervisor/render.go`). Previously peers were top-level and always injected. The template comment notes the supervisor "will not invent one," but there's no `slog.Warn` for "ring active but no neighborhood to inject into," which is an easy way to silently lose peering.

**5. (Low) `caches_changed` forces a full routing reload + peer reconcile.** `requires_routing_reload()` now returns true on `caches_changed`, so a disk-only change inside a cache triggers `build_routes` + `reload_routes` + `reconcile_peers` + a broadcast to every shard. Correct (a cache's source can change neighborhood binding), but materially more work than before for a pure disk add/remove.

### Minor / cleanliness

- `fingers_per_node.unwrap_or(0).max(1)` (`config/graph.rs`, `runtime_projection`) yields `1`, not the documented default of `100`, if `apply_defaults` wasn't run first. Works in the normal `load()` path but diverges from the documented default and from `empty_runtime_p2p()`. Prefer `.unwrap_or(100)`.
- Dead code: `compat_runtime_projection`, `CompatRuntimeProjection`, `runtime_p2p_or_default`, `first_runtime_cache`, `first_runtime_neighborhood`, `legacy_empty_projection`, `empty_runtime_p2p` are `pub`-exported (suppressing dead-code warnings) but unused; several rely on non-deterministic `HashMap::values().next()`. Recommend removing before stabilization.
- Dead error variant: `ConfigError::DuplicatePeer` is defined with a `Display` arm but never constructed (duplicates now surface as `InvalidBindingGraph`).
- Redundant validation: `validate()` calls `validate_binding_graph(cfg)` then `runtime_projection(cfg)`, which calls `validate_binding_graph` again.
- Lossy error: `validated_disk_path` returns `MissingDiskConfig(String::new())`, so the message reads `disk "": ...` with no locating info.
- Proto field numbers for removed/renumbered fields aren't `reserved`. Acceptable at a stabilization break since consumers regenerate, but worth starting that discipline for a "stabilize the schema" PR.

### Verified correct
- `StripeReq` serializes `cache_id`/`neighborhood_id` but `#[serde(skip)]`s `bypass` (local-only).
- Frontend stamps cache/neighborhood/bypass on both metadata and data requests; bypass skips fanout.
- `ChainLocalStore` lock ordering does not deadlock.
- Configmap YAML (flattened oneof) is correct for protojson; supervisor marshals to binpb so naming round-trips.
- Missing route (`""` for cache-over-backend frontends) makes `owns_locally` return true -> serves from origin, as intended.

The two I'd block on are **#1** (multi-cache CPU oversubscription) and **#2** (broken mesh-config generator). #3 and #4 are worth at least a documented guard/warning before this lands as the stable schema.

jveski:
(generated by AI)

Thanks for the detailed review. I pushed fixes for each item in separate commits.

Changes made:
- Fixed the disk CPU-slot blocker by making `DiskRegistrySet` reserve CPU slots globally across per-cache registries. Added a regression test covering two caches so they cannot both claim the same storage core.
- Fixed `gen-storage-mesh-config.sh` to emit the new strict TOML shape under `[startup.fabric.binds.tcp]`.
- Added an explicit guard for the current single fabric self-identity limitation. Configs with `local_node_id` in multiple neighborhoods now fail fast instead of advertising an inconsistent scoped identity.
- Added a supervisor warning when ring discovery is active but the config declares no neighborhoods, so discovered peers cannot be injected.
- Kept cache changes as conservative routing reloads for now and documented why: disk-only changes do not need it, but cache source/binding changes can move traffic across neighborhoods/peers.
- Cleaned up the smaller config-schema leftovers: fixed the `fingers_per_node` fallback, removed dead compat helpers and dead error variants, removed the duplicate binding-graph validation call, improved the missing disk config error, and reserved the removed `DiskSpec` fields in the proto.

Validation:
- `make unbounded-storage-test`
- `go test ./internal/storagesupervisor ./api/unbounded-storage`
- `bash -n hack/scripts/gen-storage-mesh-config.sh`

plombardi89 (approved):

PR #268, opened 2026-06-16
High level mTLS doc

jveski:
We need auth for unbounded-storage->Orca connections, and likely several more in the future (UI for example).


jveski:
I'm going to close this for now - will reopen when it becomes a priority.

PR #269, opened 2026-06-16
Unbounded-storage installer / supervisor daemonset

cchildress:
This is a K8s daemonset to handle installing and managing unbounded-storage at the host level.

The init container handles getting the base binaries and systemd units in place.
The runtime supervisor publishes config data from a K8s configmap to the unbounded-storage config file.

Still in progress is node label discovery and publishing the storage peers list.


internal/storagesupervisor/acquire.go
```diff
@@ -0,0 +1,330 @@
...
+const maxArchiveBytes = 1 << 30
+
+// httpClient is used for all release/url downloads.
+var httpClient = &http.Client{Timeout: 10 * time.Minute}
+
+// acquireAndExtract resolves the configured source into a release-layout tree
+// under stagingDir (so stagingDir/bin/unbounded-storage exists). For release
+// and url sources the tarball and its companion .sha256 are downloaded and
+// verified before extraction; a file source is extracted as-is.
+func acquireAndExtract(ctx context.Context, cfg Config, stagingDir string) error {
```
jveski, line 34: What do you think about simplifying this by distributing unbounded-storage in the installer container image? i.e. either include the tarball or its contents in the image itself.

internal/storagesupervisor/systemd.go
```diff
@@ -0,0 +1,166 @@
...
+`, cfg.Repo, cfg.Prefix, hugepageReserveCmd(cfg), configEnsureCmd(cfg), execStart)
+}
+
+// hugepageReserveCmd builds the shell one-liner used as an ExecStartPre to
+// reserve 2 MiB hugepages before the daemon starts. It mirrors the reserve
+// logic in hack/scripts/install-unbounded-storage.sh: derive a target count
+// from PoolBytes (plus headroom) unless Hugepages overrides it, attempt the
+// reservation, compact memory once if short, and fail loudly if it cannot be
+// satisfied.
+func hugepageReserveCmd(cfg Config) string {
```
jveski, line 147: nit: this faithfully reimplements the script, but fwiw there's likely a more elegant way to handle setting up hugepages using the supervisor itself

deploy/unbounded-storage-supervisor/02-configmap.yaml.tmpl
```diff
@@ -0,0 +1,64 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+---
+apiVersion: v1
+kind: ConfigMap
```
jveski, line 6: Once we have the controller wired up to generate most of the config from labels, how should values _not_ set by the controller be handled? Maybe there's a sane way to merge on top of a base config like this one, or maybe we should define a more selective "public" config schema to avoid the magic of merge logic. I haven't thought much about it either way - curious what you think

internal/storagesupervisor/acquire.go
cchildress: Maybe? The challenge there is you can't pin to a specific version as easily. You'd have to pin the supervisor container image along with it.

deploy/unbounded-storage-supervisor/02-configmap.yaml.tmpl
cchildress: I was thinking a user would update the config map when they want to change a config entry. Too simple?

jveski: Let's roll with that for now, we can figure out merging later if it's needed :+1:

internal/storagesupervisor/acquire.go
jveski: It would be nice for supervisor and unbounded-storage versions to map 1:1 so we don't have to worry about backcompat. The downside is that things get kind of strange if users need to manually bootstrap storage before k8s is up for some reason.

jveski: Only other factor I can think of is customers possibly blocking egress and breaking the tarball download

deploy/unbounded-storage-supervisor/02-configmap.yaml.tmpl
jveski: Just looked closer at how you're mapping between the configmap and the protobufs: how are you thinking about handling arrays like frontends/backends? I'm thinking it would be might to keep the literal protobufs as yaml in a configmap key for now.

internal/storagesupervisor/acquire.go
cchildress: > The downside is that things get kind of strange if users need to manually bootstrap storage before k8s is up for some reason.

I do think that's something worth preserving.

> Only other factor I can think of is customers possibly blocking egress and breaking the tarball download

Maybe the right answer is for them to have an internal mirror / repo to pull the tarball from?

jveski: That's true - we can mirror things if needed. I'm fine keeping this for now and we can figure out the details later.

deploy/unbounded-storage-supervisor/02-configmap.yaml.tmpl
cchildress: Yeah, I don't have a problem with that.

jveski (approved):
🚢

PR #305, opened 2026-06-19
KVM device passthrough

julienstroheker:
## What

Completes host KVM device passthrough by adding the missing cgroup
`DeviceAllow` grant. Today `/dev/kvm` is bind-mounted into the nspawn
container but blocked by the device cgroup controller, so KVM-accelerated
workloads (e.g. KubeVirt VMs) can't actually use it.

## Why

`DiscoverHostDevicePaths()` already binds `/dev/kvm` into the container via
`Bind=` in `nspawn.conf`. But the `DeviceAllow=` grant in
`service-override.conf` was only emitted for NVIDIA GPU devices, **not** for
host devices — so the bound device is visible but unusable inside the
container. This is effectively a latent bug: passthrough looks wired up but
silently doesn't work.

## Changes

- Emit `DeviceAllow=<dev> rwm` for host devices in the service drop-in,
  mirroring the existing GPU block.
- Tests for the new `DeviceAllow` rendering; existing golden snapshots are
  unchanged (no devices ⇒ no `DeviceAllow`).

## Notes

- KVM exposure stays **default-on** (unchanged from current behavior). An
  earlier revision gated it behind a `PassthroughKVM` config flag, but since
  the current default already exposes `/dev/kvm`, the flag would have flipped
  the default to off — a silent behavior change. Dropped it; an opt-*out*
  knob can be added later if a use case appears.
- `PrivateUsers=no` means no userns mapping is needed, so `Bind` +
  `DeviceAllow` is sufficient.
- Scoped to `/dev/kvm`; `/dev/net/tun`, `/dev/vhost-net`, and hugepages can
  follow if there's a use case.

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -47,14 +47,6 @@ type AgentConfig struct {
 	// "ghcr.io/org/repo:tag") used to bootstrap the machine rootfs.
 	// When empty the agent falls back to debootstrap.
 	OCIImage string `json:"OCIImage,omitempty"`
-
-	// PassthroughKVM, when true, exposes the host's KVM device (/dev/kvm)
-	// to the nspawn container so that KVM-accelerated workloads (e.g.
-	// KubeVirt virtual machines) can run on the node. When enabled the
-	// device is bind-mounted into the container and granted via the cgroup
-	// device controller, but only if /dev/kvm is actually present on the
-	// host; on hosts without it this is a no-op. Defaults to false.
-	PassthroughKVM bool `json:"PassthroughKVM,omitempty"`
 }
 
 // BackfillNodeName resolves and stores the Kubernetes Node name once. An
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -47,6 +47,14 @@ type AgentConfig struct {
 	// "ghcr.io/org/repo:tag") used to bootstrap the machine rootfs.
 	// When empty the agent falls back to debootstrap.
 	OCIImage string `json:"OCIImage,omitempty"`
+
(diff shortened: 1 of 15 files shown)


jveski:
@cchildress Maybe we should generalize this out to also handle passing through HCAs and block devices for storage?

pkg/agent/config/config.go
```diff
@@ -47,6 +47,14 @@ type AgentConfig struct {
...
 	// When empty the agent falls back to debootstrap.
 	OCIImage string `json:"OCIImage,omitempty"`
+
+	// PassthroughKVM, when true, exposes the host's KVM device (/dev/kvm)
+	// to the nspawn container so that KVM-accelerated workloads (e.g.
+	// KubeVirt virtual machines) can run on the node. When enabled the
+	// device is bind-mounted into the container and granted via the cgroup
+	// device controller, but only if /dev/kvm is actually present on the
+	// host; on hosts without it this is a no-op. Defaults to false.
+	PassthroughKVM bool `json:"PassthroughKVM,omitempty"`
```
bcho: oh I just realized in current impl we are already exposing the kvm device by default -- then I think we can drop this flag until we have requirement for skipping exposing kvm device. @julienstroheker wdyt?

julienstroheker: make sense, let me revert this.

julienstroheker: Dropped the flag and kept just the `DeviceAllow` fix so the existing default-on behavior is preserved and the bound device is actually usable. If we ever need an opt-out, a DisableKVM-style knob would fit better later. Updated the PR.

bcho (approved):

bcho (approved):

bcho (approved):

PR #309, opened 2026-06-22
agent: add DisableNvidia config switch

anson627:
This pull request introduces a new `DisableNvidia` option to the agent configuration, allowing users to explicitly disable NVIDIA GPU detection and runtime setup. The implementation ensures that when `DisableNvidia` is set to `true`, all related NVIDIA functionality is bypassed, even if NVIDIA devices are present on the host. The change is thoroughly tested and documented.

### DIFF
--- a/cmd/agent/internal/daemon/nodeoperator.go
+++ b/cmd/agent/internal/daemon/nodeoperator.go
@@ -136,6 +136,10 @@ func hasDrift(applied, desired *provision.AgentConfig) bool {
 		return true
 	}
 
+	if applied.DisableNvidia != desired.DisableNvidia {
+		return true
+	}
+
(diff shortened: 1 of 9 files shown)


plombardi89:
I feel like `DisableNvidia` is a bit too simplistic of a flag. What if we had a dedicated configuration block for Nvidia or really any other GPU driver wanted to support, for example:

```
{
  "gpu": {
    "kind": "amd", "nvidia", "intel", "none" // none used when someone wants to do their own thing. `none` would also the be the default if `gpu` block is null.
    "nvidia": { /* nvidia customization options */ }
    "amd": { /* amd customization options */ }
    "intel": ... same pattern ...
  }
}
```

PR #315, opened 2026-06-22
storage: In-place benchmark tooling

jveski:
- Node annotation used to allocate block devices to unbounded-storage
- Benchmark harness driven by k8s annotations, makes it easy to generate synthetic load in real clusters
- Configurable write depth for RDMA CQs


api/unbounded-storage/config.proto
```diff
@@ -306,4 +306,17 @@ message LoadgenFrontendConfig {
...
+  optional uint64 fixed_object_size_bytes = 6;
+
+  // Only issue reads whose generated data stripes route to a peer rather than
+  // being owned locally. Requires fixed_object_size_bytes and a frontend source
+  // that resolves through a neighborhood.
+  optional bool require_remote_peer = 7;
+
+  // Suppress frontend_request metrics for this many successful operations per
+  // worker before reporting benchmark traffic. Useful for cache warmup.
+  optional uint64 warmup_operations = 8;
```
jveski, line 348: why uint64?

deploy/unbounded-storage-supervisor/02-configmap.yaml.tmpl
```diff
@@ -25,6 +25,40 @@ metadata:
...
+# Storage benchmarks are also supervisor-injected from Node annotations. They
+# run between a source node and a target node. Both nodes must be visible to
+# this supervisor's node watch, so they must carry the storage-ring label key
+# even if the TCP ring itself is not active. The source node carries:
+#   unbounded-cloud.io/storage-benchmark.scenario: rdma-loadgen | rdma-cache-miss | tcp-loadgen | tcp-cache-miss
+#   unbounded-cloud.io/storage-benchmark.target-node: <target-node-name>
+# Both source and target nodes carry either their provider-native RDMA endpoint:
+#   unbounded-cloud.io/storage-rdma.addr: hex:<fi_getname-bytes>
+# or their libfabric TCP socket endpoint:
+#   unbounded-cloud.io/storage-tcp.addr: <host>:<port>
```
jveski: Let's use the Node's internal IP, with a configurable port defaulting to the default tcp listen port.

Makefile
```diff
@@ -245,6 +245,7 @@ help: ## Show this help
 	@echo "  unbounded-storage | unbounded-storage-build  Build unbounded-storage (with/without test)"
 	@echo "  UNBOUNDED_STORAGE_PROFILING=1     Set on any build/push to compile in the SIGUSR1 CPU profiler"
 	@echo "  unbounded-storage-smoke          Run the end-to-end smoke test (uses sudo)"
+	@echo "  unbounded-storage-supervisor-smoke-kind  Run the storage supervisor kind smoke test"
```
jveski: Let's completely remove the new smoke tests added in this branch for now.

jveski (commented):
Trying a new thing - reviewing my own PR and asking an agent to make the fixes

internal/storagesupervisor/benchmark.go
```diff
@@ -0,0 +1,700 @@
...
+		cfg.WarmupOperations = proto.Uint64(bench.effectiveWarmupOperations())
+	}
+
+	return cfg
+}
+
+func benchmarkCache(bench rdmaLoadgenBenchmark, cacheName, neighborhoodName string) *storageconfig.CacheSpec {
+	cache := &storageconfig.CacheSpec{Name: cacheName, Source: neighborhoodName}
+	if !bench.runLoadgen {
+		cache.Disks = []*storageconfig.DiskSpec{
```
jveski: The disk annotation should be used to configure available disks globally, not just for the benchmarks.

api/unbounded-storage/config.proto
jveski: (AI AGENT RESPONSE)

I kept this as `uint64` because it is an operation counter in the same domain as `object_count` and the internal loadgen sequence number. The daemon compares each generated operation sequence against `warmup_operations`, so keeping it `uint64` avoids a smaller API limit for long warmup runs.

deploy/unbounded-storage-supervisor/02-configmap.yaml.tmpl
jveski: (AI AGENT RESPONSE)

Updated TCP benchmark addressing in commit `ef295cc`: TCP scenarios now use each Node's InternalIP plus the configured `startup.fabric.tcp.addr` port by default, with `unbounded-cloud.io/storage-tcp.port` as a per-node override.

Makefile
jveski: (AI AGENT RESPONSE)

Removed the new supervisor kind smoke workflow and harness in commit `bbc7dd7`.

internal/storagesupervisor/benchmark.go
jveski: (AI AGENT RESPONSE)

Updated this in commit `bffc11e`: disk annotations now compute a node-wide supervisor disk overlay, and benchmark caches consume that shared overlay rather than carrying benchmark-specific disk fields.

hack/smoke-storage-supervisor-kind.py
```diff
@@ -0,0 +1,661 @@
...
+
+
+def tool(name: str) -> str:
+    path = shutil.which(name)
+    if not path:
+        die(f"required tool not found in PATH: {name}")
+    return path
+
+
+def container_engine() -> str:
```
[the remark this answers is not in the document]
jveski: (AI AGENT RESPONSE)

The supervisor kind smoke test file was removed in commit `bbc7dd7`, so this function no longer exists.

PR #328, opened 2026-06-23
ci: sign releases in tag context to fix cosign verification

plombardi89:
## Problem

The deploy pipeline (`release-upgrade.yaml`) verifies the release manifests tarball with a cosign **keyless** signature and requires the signing certificate identity to match:

```
^https://github.com/Azure/unbounded/.github/workflows/release.yaml@refs/tags/<tag>$
```

With keyless signing there is no key; trust is anchored in that identity, which encodes the workflow file **and the git ref it ran on**. The identity only ends in `@refs/tags/<tag>` when `release.yaml` runs because a **tag was pushed**.

But releases were being cut via `release.yaml`'s `workflow_dispatch` "bump" path. That run executes on `main`, and creating the tag mid-run does not change the ref the run is on. So every keyless signature was stamped `release.yaml@refs/heads/main`. The tag it pushed used the default `GITHUB_TOKEN`, which GitHub deliberately prevents from triggering further workflows, so no tag-context run ever happened.

The result is the deploy failing verification:

```
Error: failed to verify certificate identity: no matching CertificateIdentity found,
expected SAN value to match regex "...release.yaml@refs/tags/v0.1.16$",
got "...release.yaml@refs/heads/main"
```

This is a structural mismatch: the release supports a dispatch path that signs as `@refs/heads/main`, but the deploy is built to trust only `@refs/tags/<tag>`.

## Solution

Split the release into two phases so signing **always** happens in tag context:

- **`release-prepare.yaml` (new, Phase 1):** a `workflow_dispatch` that only computes the next version and pushes the tag. It pushes using an **SSH deploy key** (`RELEASE_TAG_DEPLOY_KEY`) rather than `GITHUB_TOKEN`, so the tag push is allowed to trigger Phase 2.
- **`release.yaml` (Phase 2):** now tag-driven only (`on: push: tags`). Removed the in-run tag creation; the build/sign jobs run from the pushed tag.

Because Phase 2 is triggered by the tag push and checks out the tag, every keyless signature now carries the identity `release.yaml@refs/tags/<tag>`. The deploy's verification passes **with no change to the verifier** and full provenance is preserved (version-pinned, immutable identity).

## How to cut a release (developer workflow)

The recommended way to cut a release is the **unbounded release prepare** workflow. (Pushing a tag by hand still works as a break-glass path; see below.)

1. Make sure `main` is at the commit you want to release (the tag is always cut from `main`, regardless of which ref you launch the workflow from).
2. In the GitHub UI, go to **Actions -> unbounded release prepare -> Run workflow**, and pick a `mode`:
   - **`release`** + `bump` (`patch`/`minor`/`major`) - cut a normal release. The next version is computed by bumping the latest **final** tag. Example: latest final `v0.2.3`, `bump=patch` -> `v0.2.4`.
   - **`prerelease`** + `bump` + `pre` - cut a prerelease. Example: latest final `v0.2.3`, `bump=minor`, `pre=rc.1` -> `v0.3.0-rc.1`. Iterate the same train by re-running with the **same** `bump` and an incremented `pre` (`rc.1` -> `rc.2`).
   - **`promote`** - finalize an in-flight prerelease to its release version, with **no** bump. By default it promotes the highest prerelease train (e.g. `v0.3.0-rc.2` -> `v0.3.0`). Set the optional `version` input (e.g. `v0.3.0`) to promote a specific train when several are in flight; that `vX.Y.Z-*` train must exist or the run fails.
3. The workflow pushes the tag, which automatically triggers `release.yaml` to build, sign, and create a **draft** GitHub release for that tag.
4. Verify the draft release and its artifacts, then **publish manually** (GitHub UI or `gh release edit <tag> --draft=false`). The tag push also auto-triggers the deploy to the `stable` soak cluster.

Notes:
- Prerelease status is derived from the tag (a `-` suffix marks a prerelease); there is no separate prerelease toggle.
- If a tag gets pushed but no `release.yaml` run starts (e.g. a deploy-key/policy issue), delete the tag (`git push origin :refs/tags/<tag>`) and re-run **release prepare**. There is intentionally no manual `release.yaml` trigger: a dispatched run would sign as `@refs/heads/main` and fail verification.

### Cutting a release by tag (supported but discouraged)

Manually pushing a `v*` tag still triggers `release.yaml` and produces a correct release: because the tag is pushed with your own credentials (not the bot `GITHUB_TOKEN`), the run executes in tag context and signs as `release.yaml@refs/tags/<tag>`, so verification and the auto-deploy work exactly as via **release prepare**.

```
git tag v0.3.0 && git push origin v0.3.0
```

It is discouraged because it bypasses the guardrails that **release prepare** provides:
- **The `ref: main` guard.** A hand-pushed tag can point at any commit on any branch, so you can accidentally release un-merged or feature-branch code. The signature still verifies (the identity is tag-based), so nothing downstream catches it.
- **Version computation and guards** (mode-based release/prerelease/promote, bump-from-latest-final, train iteration, duplicate/typo checks). With a manual tag you pick the version string yourself; only the basic `vX.Y.Z[-suffix]` format check in `release.yaml` applies.

Prefer **release prepare** for normal releases; reserve manual tagging for break-glass situations (e.g. the prepare workflow is unavailable).

## Also in this change
- Dropped the dry-run paths from `release.yaml` (build validation is covered by CI).
- Releases are always left as drafts; publishing remains a manual step.
- `release.yaml` no longer publishes `:latest` image tags. Nothing in the shipped/supported path consumes them (rendered manifests, the `kubectl-unbounded` plugin's embedded manifests, and the integration deploy all pin to `:vX.Y.Z`), and a mutable `:latest` is a footgun now that prereleases flow through the same pipeline (they would silently repoint `:latest` to an RC). Releases publish only the immutable `:vX.Y.Z` tag; existing `:latest` tags in the registry are left as-is.

## Setup required (already completed)
- An SSH deploy key with write access is added to the repo, and its private key is stored as the `RELEASE_TAG_DEPLOY_KEY` Actions secret.

## Behavior change to note
Real releases are now `push` events, so the deploy workflow auto-fires after each release (previously it was run manually).

## Verification
- `actionlint` passes on both workflows.
- Both files parse as valid YAML; no remaining dry-run/dispatch input references.
- The `release-prepare` version logic was traced across all modes against a scratch repo (release/prerelease/promote, the `version` override, and multi-train cases).
- End-to-end proof requires a live run: dispatch **release prepare**, confirm `release.yaml` runs from the tag, a signature shows `...release.yaml@refs/tags/<tag>`, and the deploy's cosign verify step passes.

### DIFF
--- a/.github/workflows/release-prepare.yaml
+++ b/.github/workflows/release-prepare.yaml
@@ -19,13 +19,38 @@ name: unbounded release prepare
 # triggers for tags pushed with GITHUB_TOKEN (loop prevention); a deploy-key
 # push is exempt, so it triggers release.yaml's `on: push: tags` in the
 # correct tag context.
+#
+# Modes (the `mode` input):
+#   release     Bump the latest FINAL tag by `bump` and cut vX.Y.Z.
+#   prerelease  Bump the latest FINAL tag by `bump` and cut vX.Y.Z-<pre>
+#               (e.g. pre=rc.1). Iterate a train by re-running with the same
+#               `bump` and an incremented `pre` (rc.1 -> rc.2 -> ...).
+#   promote     Finalize an in-flight prerelease to its release version, with
+#               NO bump. By default this promotes the highest prerelease
+#               train; set `version` (e.g. v0.2.0) to promote a specific one.
+#
+# Recovery: if a tag is pushed but no release.yaml run is triggered (for
+# example a misconfigured deploy key or disabled Actions), delete the tag
+# (`git push origin :refs/tags/<tag>`) and re-run this workflow. There is
+# deliberately NO workflow_dispatch fallback on release.yaml: a dispatched
+# run would sign as `@refs/heads/main` and fail deploy verification, which is
+# the exact problem this two-phase split exists to avoid.
 # ---------------------------------------------------------------------------
 
 on:
   workflow_dispatch:
     inputs:
+      mode:
+        description: "release: cut vX.Y.Z | prerelease: cut vX.Y.Z-<pre> | promote: finalize a prerelease"
+        required: true
+        type: choice
+        options:
+          - release
+          - prerelease
+          - promote
+        default: release
       bump:
-        description: "Version bump level"
+        description: "Version bump level (release/prerelease)"
         required: true
         type: choice
         options:
@@ -34,7 +59,12 @@ on:
           - major
         default: patch
       pre:
-        description: "Optional prerelease suffix (e.g. rc.1); leave blank for a normal release"
+        description: "Prerelease suffix for prerelease mode (e.g. rc.1)"
+        required: false
+        type: string
+        default: ""
+      version:
+        description: "promote only: explicit final version to cut (e.g. v0.2.0); blank = highest in-flight prerelease"
         required: false
         type: string
         default: ""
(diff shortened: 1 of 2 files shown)


plombardi89:
Findings from reviewing the release workflow changes:

1. `.github/workflows/release-prepare.yaml:74-93` mishandles prerelease trains. If the latest tag is `v0.2.0-rc.1`, another `patch` with `pre=rc.2` produces `v0.2.1-rc.2`, and a final blank `pre` produces `v0.2.1` instead of promoting to `v0.2.0`. That can make RC iteration/finalization publish the wrong version.

2. `.github/workflows/release-prepare.yaml:56-61` and `:108` tag whatever ref the manual dispatch checked out. If someone runs the prepare workflow from a feature branch, it releases that branch, and the tag-context signatures will still verify. If releases must come from `main`, this needs a guard or an explicit `ref: main` checkout.

3. `.github/workflows/release-prepare.yaml:61` depends on `RELEASE_TAG_DEPLOY_KEY` being configured as a write-enabled deploy key, and `.github/workflows/release.yaml:35-38` removes the manual fallback. If that key/policy does not trigger Actions on tag push, the tag can exist without a release run; recovery then requires deleting/re-pushing the tag or restoring a manual trigger.

The core cosign fix looks directionally sound: prior tag-push release runs show `headBranch` as the bare tag, so `release-upgrade.yaml` should still resolve the tag and verify `release.yaml@refs/tags/<tag>`.

plombardi89:
Thanks for the review. Addressed all three findings in 1e5c796 (changes confined to `release-prepare.yaml`; `release.yaml`/`release-upgrade.yaml` unchanged).

**1. Prerelease trains (was: always re-bumps, finalize produces wrong version).**
Replaced `bump`+`pre` with an explicit `mode` input `{release, prerelease, promote}`, and the bump base is now the latest **final** tag (excludes prerelease tags, which also sidesteps git's version-sort ranking `v0.2.0-rc.1` above `v0.2.0`).
- `prerelease` + same `bump` + incremented `pre` iterates a train: `v0.2.0-rc.1` -> `v0.2.0-rc.2`.
- `promote` finalizes with **no bump**: `v0.2.0-rc.2` -> `v0.2.0`.
- Added an optional `version` override for `promote` to finalize a specific in-flight train when several exist (guarded: a `vX.Y.Z-*` train must exist, else it errors).

**2. Tags whatever ref was dispatched.**
Checkout is now pinned to `ref: main`, so the tag is always cut from the default branch regardless of which ref the manual dispatch selected.

**3. Deploy-key dependency / orphan-tag recovery.**
Added a preflight that fails **before** any tag is created if `RELEASE_TAG_DEPLOY_KEY` is missing or `git ls-remote origin` can't authenticate. The header documents the recovery path (delete the tag, re-run prepare) and notes that a `workflow_dispatch` fallback on `release.yaml` is intentionally omitted because a dispatched run would sign as `@refs/heads/main` and fail verification, the exact issue this split fixes.

Verification: `actionlint` clean, YAML parses, and I traced all modes (including the `version` override and multi-train cases) against a scratch repo, e.g. finals `v0.1.5` + in-flight `v0.2.0-rc.2`/`v0.3.0-rc.1`: `promote` (blank) -> `v0.3.0`, `promote version=v0.2.0` -> `v0.2.0`, `promote version=v0.4.0` -> error.

jveski (dismissed):

jveski (approved):

PR #332, opened 2026-06-23
storage: fix arm64 c_char FFI buffer

jveski:
## Summary
- fix the unbounded-storage sockaddr formatting buffers to use platform c_char
- avoid assuming C char is signed, which breaks the arm64 release build where c_char is u8

## Root cause
The release workflow failed in storage-binaries (arm64) during make unbounded-storage-build because ub_fi_format_sockaddr expects *mut c_char, but cm.rs passed buffers typed as i8. On aarch64-unknown-linux-gnu, Rust c_char is u8, so the pointer types do not match.


plombardi89 (commented):
Any idea how it snuck past CI in the fist place?

jveski:
We only test against x86 and this only fails on arm :skull:

plombardi89 (approved):

PR #339, opened 2026-06-24
Mount storage and IB devices into the nspawn environment

cchildress:
Mount storage and IB devices into the nspawn environment for unbounded-storage.

This allows unbounded-storage to run inside our isolated environment, but it still has control over local disks and the ability to use the IB devices.


jveski:
Looks good. Only question is about how to handle disk swaps. If the machine is rebooted will the agent update the nspawn mounts? I don't have enough context on how this code is called to know.

jveski (approved):

PR #344, opened 2026-06-25
agent: expose AMD GPU devices in nspawn

bcho:
## Summary
- discover AMD GPU device nodes on the host
- bind-mount /dev/kfd and DRM card/render nodes into nspawn
- grant matching DeviceAllow access so the AMD Kubernetes device plugin can detect devices inside the machine

## Tests
- go test ./pkg/agent/goalstates ./pkg/agent/phases/rootfs

### DIFF
--- a/pkg/agent/goalstates/amd.go
+++ b/pkg/agent/goalstates/amd.go
@@ -10,9 +10,7 @@ import (
 	"strings"
 )
 
-const (
-	amdKFDDevicePath = "/dev/kfd"
-)
+const amdKFDDevicePath = "/dev/kfd"
 
 // AMDHost aggregates AMD GPU host state discovered at agent startup.
 type AMDHost struct {
--- a/pkg/agent/goalstates/amd.go
+++ b/pkg/agent/goalstates/amd.go
@@ -68,6 +68,12 @@ func discoverAMDDevicesAt(kfdPath, driPath string) []string {
 			}
 
 			name := e.Name()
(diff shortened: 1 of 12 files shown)


jwilder (dismissed):
Can we update docs to reflect we detect AMD gpus now?

pkg/agent/goalstates/amd.go
```diff
@@ -0,0 +1,94 @@
...
+package goalstates
+
+import (
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+)
+
+const amdKFDDevicePath = "/dev/kfd"
```
anson627, line 13: This might be too broad for GPU nodes with multiple DRM producers including unrelated host devices inside the nspawn machine and bypasses the intended device boundary. For example, if /dev/kfd exists, we bind-mount and DeviceAllow every /dev/kfd/card*, including iGPU/BMC/NVIDIA/display DRM nodes that are not AMD devices

anson627 (dismissed):

pkg/agent/goalstates/amd.go
bcho: this might not that easy to restrict as requires to check the vendor information about the devices. I am also afraid there will be some corner cases. On the other hand, having different GPU devices from the same node should be rare. I will keep a TODO for this and see if we can further lock it down in the future.

cchildress (approved):

PR #381, opened 2026-06-30
metalman: fix concurrent operations

jveski:
Currently only a single metalman `MachineOperation` is processed per site. This change adds conflict detection so it's safe to run concurrent provisioning operations.


imiller31 (approved):
LGTM

PR #386, opened 2026-06-30
[unbounded-system] unbounded-operator: Site API, component reconcile, and released-cluster migration

plombardi89:
## Problem

Unbounded has no top-level object representing a location and the components running there, and no controller that turns declared intent into workloads. Today:

- Installation is imperative - a chain of `kubectl unbounded` steps that download and apply CNI/machina manifests by hand.
- `Site` lives in the networking API group (`net.unbounded-cloud.io/v1alpha1`) even though it is conceptually a cluster-wide location.
- Components are split across namespaces (`unbounded-kube` for machina/metalman/storage, `unbounded-net` for net) and wired by hand.
- Released clusters already run that split layout, so the new model needs a safe automated migration. Kubernetes namespaces cannot be renamed in place.

## What this PR changes

- Moves `Site` to the cluster-scoped `unbounded-cloud.io/v1alpha3` API and makes it the top-level declarative object for a location.
- Adds `unbounded-operator`, installed in `unbounded-system`, to reconcile `Site` intent into the component workloads.
- Consolidates operator-managed components into the unified `unbounded-system` namespace.
- Adds an always-on, leader-elected legacy reaper that migrates v0.1.19-style split-namespace clusters onto the new model.
- Retargets net, kubectl, docs, release manifests, and CI/e2e plumbing to the new model.

## Behavioral changes

- **`Site` API group changes.** New Sites are `sites.unbounded-cloud.io` (`unbounded-cloud.io/v1alpha3`). The legacy `sites.net.unbounded-cloud.io` CRD is translated and then removed during migration.
- **Components are declared under `Site.spec.components`.** `machina`, `metalman`, and `storage` each use `{ enabled }`; metalman additionally has `dhcpAutoInterface`. Net is not a component because it is a cluster singleton required to interpret every Site's networking spec.
- **Net and machina are persistent singletons.** The operator deploys net when at least one Site exists and machina when any Site enables it, but it does not auto-delete either singleton when Sites are removed or machina is disabled everywhere. Net is the dataplane, and deleting net-node can cause a cluster-wide outage; singleton removal is left to a future explicit uninstall flow.
- **Per-site components are declarative.** Disabling `storage` or `metalman` on a Site deletes that Site's storage/metalman workload. Deleting a Site garbage-collects its per-site storage/metalman resources via ownerReferences.
- **Storage config is ConfigMap-owned.** `Site.spec.components.storage.config` is not part of the API. Storage config lives in `unbounded-storage-config-<site>` in `unbounded-system`: the operator creates it from the embedded default if absent, adopts/preserves it if present, and mounts it into the per-site storage DaemonSet. Operators edit the ConfigMap directly.
- **Node site label is changing.** `unbounded-cloud.io/site` is the canonical Node site label. `net.unbounded-cloud.io/site` is deprecated but dual-written and read as a fallback during this transition. Storage and metalman use node affinity matching either key so they schedule correctly before upgraded net has converged labels.
- **hostNetwork controllers roll old-before-new.** net-controller and per-site metalman use `RollingUpdate` with `maxSurge: 0` so a rollout frees host ports before the replacement pod starts. A bad new image can create a zero-available window; roll back failed rollouts and alert on `AvailableReplicas == 0`.
- **`kubectl unbounded install` is the install entrypoint.** Imperative per-component install commands are removed. `kubectl unbounded site init` writes declarative Sites; `--enable-storage` applies to the remote Site.
- **Install namespace is authoritative.** `--namespace` plus injected `POD_NAMESPACE` drives where operator-managed workloads and migrated state go. Custom namespaces are supported by manifest retargeting but are not yet covered by the faithful upgrade e2e.

## Component model

`unbounded-operator` reconciles components as follows:

- **net** - cluster singleton, deployed whenever at least one Site exists. It reads every Site's networking spec and writes Node labels/CIDRs.
- **machina** - cluster singleton, deployed when any Site enables `components.machina`.
- **metalman** - per-Site Deployment plus shared RBAC. It runs on Nodes carrying either the canonical or deprecated site label during the label transition.
- **storage** - per-Site DaemonSet, per-Site ConfigMap (`unbounded-storage-config-<site>`), and shared RBAC. The DaemonSet is node-affined to either site label during the label transition.

Component images/versions come from the operator release (version-matched embedded manifests). The operator config is reduced to `{ MetalmanImage, APIServerEndpoint }` plus namespace/reaper flags.

## Legacy migration / reaper

The reaper runs continuously under leader election. Each idempotent pass:

1. Translates legacy net-group Sites to machina-group Sites, preserving the networking spec and inferring enabled components from legacy workloads.
2. Copies non-regenerable Secrets and `machina-config`; upserts `machina-config` so migrated config wins over default reconcile content.
3. Copies legacy shared `unbounded-storage-config` into `unbounded-storage-config-<site>` for every storage-enabled translated Site. The operator later adopts those ConfigMaps without changing data.
4. Copies each Machine cloud-init `userDataConfigMapRef` ConfigMap and rewrites the ref namespace.
5. Rewrites cluster-scoped secret refs on `Machine` and `MachineOperationCredential` from legacy namespaces to `unbounded-system`.
6. Reaps legacy component resources after target readiness gates pass. Storage waits for every storage-enabled Site's per-site DaemonSet; net is reaped last and gated on target creation rather than readiness because old/new net-node cannot share host ports.
7. Clears the old net Site protection finalizer, deletes the legacy net Site CRD, and deletes drained legacy namespaces.

## Site label deprecation details

- Net writes both `unbounded-cloud.io/site` and `net.unbounded-cloud.io/site` on Nodes and reads canonical-first with deprecated fallback.
- The net node-field `ValidatingAdmissionPolicy` allows the net service account to write the canonical label.
- `SiteNodeSlice` ownerReferences point at the new `unbounded-cloud.io/v1alpha3` Site GVK.
- Managed kube-proxy uses the canonical label and recreates its DaemonSet when the immutable selector changes from the deprecated label to the canonical label.

## Release / build / CI

- The release image flow now includes `unbounded-operator` (signing + SBOM).
- Storage supervisor image tags default to `$(VERSION)`; guard tests prevent embedded `:latest` image tags.
- Operator image builds accept component-image build args so e2e can point embedded manifests at a local registry.
- Release manifests include the operator manifests.

## Tests and verification

- Unit tests cover API registration/deepcopy, reconcile mutations, ownerRefs, storage config create/adopt/preserve, migration/reaper behavior, RBAC, site-label dual-write/fallback, kube-proxy selector recreation, and release/image guards.
- In-process reaper e2e (`go test -tags=e2e -run TestOperatorReaperMigration ./e2e/operator/...`) runs in the per-PR workflow and covers storage/metalman migration logic with staged workloads.
- Faithful released-version upgrade e2e (`python3 hack/operator-upgrade-e2e/e2e.py all`) stands up kind, installs v0.1.19, upgrades to this tree, and verifies real migration of net + machina, Node label dual-write, and a clean net-controller rollout restart. Storage (RDMA) and metalman (PXE) cannot run on vanilla kind, so their scheduling paths are unit/in-process-e2e covered.

Latest local validation:

- `make fmt`
- `make lint`
- `make build`
- `make test` (103 packages, 0 failures)
- `go build -tags=e2e ./e2e/...`
- `python3 hack/operator-upgrade-e2e/e2e.py all` (faithful v0.1.19 -> operator upgrade e2e, kind)

## How to review this PR

Five commits, reviewable in order:

1. **`Introduce unbounded-operator ...`** - Site API move, operator, component reconcile, net/kubectl retargeting.
2. **`Migrate existing clusters ...`** - reaper, migration RBAC, and e2e/regression tests.
3. **`[unbounded-system] Address PR386 review ...`** - initial review fixes, site-label deprecation, migration hardening, and hostNetwork rollout fix.
4. **`[unbounded-system] Harden site lifecycle ...`** - ownerReferences for per-site resources, safe singleton lifecycle, and deprecated-label scheduling bridge.
5. **`[unbounded-system] Make storage config ConfigMap-owned`** - removes `Site.spec.components.storage.config`; storage config is created/adopted as `unbounded-storage-config-<site>` and migration copies legacy shared config per Site.

Suggested reading order: **Site API -> Operator reconcile -> Migration/reaper -> Net site-label deprecation -> kubectl -> Build/release -> Tests/docs.**

## Try it locally

Full faithful upgrade e2e (build images, install v0.1.19, upgrade, verify, cleanup):

```bash
python3 hack/operator-upgrade-e2e/e2e.py all
```

Step-by-step with the cluster kept for inspection:

```bash
python3 hack/operator-upgrade-e2e/e2e.py setup        --keep-cluster
python3 hack/operator-upgrade-e2e/e2e.py build-images --keep-cluster
python3 hack/operator-upgrade-e2e/e2e.py install-old  --keep-cluster
python3 hack/operator-upgrade-e2e/e2e.py upgrade      --keep-cluster
python3 hack/operator-upgrade-e2e/e2e.py verify       --keep-cluster
python3 hack/operator-upgrade-e2e/e2e.py cleanup
```

Faster in-process reaper smoke:

```bash
go test -tags=e2e -run TestOperatorReaperMigration ./e2e/operator/...
```

## Scope / follow-ups

- Targets `feature/unbounded-system`, not `main`.
- The net cutover has a brief accepted data-plane gap because old/new net-node cannot share host ports. A zero-gap cutover would require distinct migration health ports and is out of scope.
- A future release should complete the site-label deprecation by removing the deprecated `net.unbounded-cloud.io/site` write/fallback and the transitional storage/metalman OR affinity.
- A future explicit uninstall flow should handle safe removal of net/machina singletons.


plombardi89:
Code/architecture review findings, focused on migration sharp edges:

1. High: `SiteNodeSlice` ownerRefs still point at the old Site API.  
`internal/net/controller/site_controller.go:1204-1208`  
Owner refs use `net.unbounded-cloud.io/v1alpha1` `Site`, but Site now lives in `unbounded-cloud.io/v1alpha3`. This can leave slices orphaned or referencing a removed GVK after migration. Please switch to `unboundedv1alpha3.GroupVersion.String()` and add a regression test.

2. High: disabling/deleting components does not remove workloads.  
`internal/operator/reconciler.go:73-77`, `123-127`, `153-160`, `187-188`  
The reconciler reports `"no sites"`/`"disabled"` but does not delete singleton net/machina or per-site metalman/storage resources. This makes `spec.components.*.enabled` non-declarative and can leave hostNetwork/storage workloads running. Please add cleanup/finalizer paths.

3. High: metalman uses the Machine site label as a Node selector.  
`internal/operator/reconciler.go:428-430`  
Nodes are labeled by net with `net.unbounded-cloud.io/site`, while metalman selects `unbounded-cloud.io/site`. I don't see a normal path adding that label to Nodes, so per-site metalman Deployments can remain Pending. Please use `siteLabelKey`.

4. High: storage migration can reap legacy storage after only one target DaemonSet is ready.  
`internal/operator/migrate.go:705-728`  
`storageTargetsReady` only requires at least one `unbounded-storage-supervisor-*` DaemonSet. Multi-site clusters can lose the legacy supervisor before every storage-enabled Site has a replacement. Please require the exact per-site DaemonSet for every storage-enabled translated Site.

5. High: Machine cloud-init ConfigMap refs are not migrated.  
`api/machina/v1alpha3/machine_types.go:294-299`, `internal/operator/migrate.go:159-172`, `534-557`, `781-789`  
The reaper rewrites secret refs but not `Machine.spec.pxe.cloudInit.userDataConfigMapRef`; those ConfigMaps can be deleted with `unbounded-kube`. Please copy/rewrite these refs or block namespace deletion.

6. High: migrated ConfigMaps can be overwritten by reconcile defaults.  
`cmd/unbounded-operator/main.go:110-112`, `internal/operator/migrate.go:534-557`, `internal/operator/reconciler.go:279-281`, `296-305`, `332-335`  
The reaper preserves `machina-config`/`unbounded-storage-config`, but the reconciler force-applies embedded defaults. Storage config is also one shared ConfigMap despite being per-Site in the API. Please preserve migrated data and clarify/fix storage config scoping.

7. Medium: `site init --enable-storage` only enables storage on the cluster Site.  
`cmd/kubectl-unbounded/app/site_init.go:128-155`  
Remote Site config does not set `EnableStorage`, even though storage is reconciled per Site. Please either enable it on the remote Site too or make the flag scope explicit.

8. Medium: custom install namespace support is inconsistent during migration.  
`cmd/kubectl-unbounded/app/install.go:76`, `160-180`; `cmd/unbounded-operator/main.go:106-109`  
`--namespace` says it controls operator and components, but the reaper target is still derived as `unbounded.SystemNamespace()` in the operator pod. Please make one namespace source authoritative or remove/clarify custom namespace support for this migration.

9. Medium: release workflow does not appear to publish the new operator image.  
`.github/workflows/release.yaml:448-474`; `Makefile:66`, `1009-1018`  
Release manifests reference `ghcr.io/azure/unbounded-operator:<tag>`, but the release image matrix does not include `unbounded-operator`. Fresh installs can ImagePullBackOff. Please add it to the signed/SBOM image flow.

10. Medium: storage supervisor manifests render `:latest`.  
`Makefile:102-103`, `633-640`; `deploy/unbounded-storage-supervisor/rendered/04-daemonset.yaml:43`, `68`  
Release rendering can still embed `ghcr.io/azure/unbounded-storage-supervisor:latest`, which breaks version-matched operator-managed components. Please default to `$(VERSION)` and add an operator build override if needed for e2e.

api/machina/v1alpha3/site_types.go
```diff
@@ -0,0 +1,178 @@
...
+	Components SiteComponents `json:"components,omitempty"`
+}
+
+// SiteComponents declares optional components managed by the unbounded operator.
+// Networking (unbounded-net) is not a component: it is a cluster singleton the
+// operator deploys whenever at least one Site exists.
+type SiteComponents struct {
+	// Machina configures the machina controller for this site.
+	// +optional
+	Machina *SiteComponentSpec `json:"machina,omitempty"`
```
jveski: nit: `MachinaComponentSpec` in case we need to add fields in the future

api/machina/v1alpha3/site_types.go
```diff
@@ -0,0 +1,178 @@
...
+	// +optional
+	NodeCount int `json:"nodeCount,omitempty"`
+
+	// SliceCount is the number of SiteNodeSlice objects for this site.
+	// +optional
+	SliceCount int `json:"sliceCount,omitempty"`
+
+	// Components reports the last observed state of site components.
+	// +optional
+	Components map[string]SiteComponentStatus `json:"components,omitempty"`
```
jveski: nit: if these were conditions it would be possible to `kubectl wait` the Site object which might be kind of nice

jveski (approved):

api/machina/v1alpha3/site_types.go
plombardi89: Done in 9f7c4037: introduced a dedicated `MachinaComponentSpec` (embedding `SiteComponentSpec`, mirroring `MetalmanComponentSpec`/`StorageComponentSpec`) so machina-specific fields can be added later without an API break.

api/machina/v1alpha3/site_types.go
plombardi89: Done in 9f7c4037: replaced `status.components` (map of a custom `SiteComponentStatus`) with a standard `[]metav1.Condition`. The operator now publishes one condition per component (`NetReady`, `MachinaReady`, `MetalmanReady`, `StorageReady`), so `kubectl wait --for=condition=NetReady site/<name>` works.

PR #389, opened 2026-07-01
metalman: separate netboot image

jveski:
- Netboot targets a different OCI image than the target "machine" image. Decouples the boot image from the installer image.
- Adds support for ARM64 netbooting
- Removes the agent from the host image. PXE machines can now specify any agent image.


cmd/metalman/README.md
```diff
@@ -19,7 +19,7 @@ metadata:
   name: node-01
 spec:
   pxe:
-    image: ghcr.io/azure/images/host-ubuntu2404:v1
+    image: ghcr.io/azure/host-ubuntu2404:v1
```
imiller31, line 22: was this incorrect before?

imiller31 (approved):

cmd/metalman/README.md
jveski: yeah apparently it was :upside_down_face:

PR #397, opened 2026-07-02
metalman: uefi http boot support

jveski:
Adds support for Redfish controlled UEFI HTTP boot in metalman. Currently there is not any e2e coverage because the Sushy emulator doesn't support it.


imiller31 (approved):
LGTM

PR #440, opened 2026-07-10
storage: simplify and harden the cache engine

jveski:
- simplify admission to a second-touch doorkeeper and singleflight to a non-blocking per-key lease
- carry key and size metadata in LRU residents so eviction can commit deletes before reclaiming storage without a reverse map
- reduce B-tree commit and checksum allocations by updating the mirror incrementally and verifying pages in place
- remove unused engine configuration, block-device APIs, and dependencies


jveski:

- **High - concurrent eviction can delete a newer overwrite and free the wrong range** (`cmd/unbounded-storage/src/storage/engine.rs:415`, `cmd/unbounded-storage/src/storage/engine.rs:439-443`): the new folded `Resident` record is captured by `sweep`, but eviction submits only its key and never verifies that the committed B-tree value still has `resident.lba`. A same-key overwrite can commit after the sweep; the unconditional delete then removes that newer mapping, while the post-commit `free_range(resident.lba, ...)` can target a range the overwrite path already retired and the allocator may have reassigned. This can turn a cache race into lost entries or cross-entry corruption. Make deletion conditional on `(key, expected_lba)` and free only when that exact mapping was removed.

cchildress (approved):

PR #448, opened 2026-07-10
storage: supervise startup and config application

jveski:
- gate shard serving on phase-B activation and initial disk publication
- precompute and share loaded config projections and route snapshots
- preserve live resources on failed reloads and surface disk and peer reconciliation failures
- supervise shard termination and fail coherently on indeterminate config application


jveski:

- **Blocker:** `cmd/unbounded-storage/src/main.rs:922-972` loses failed forced frontend rebuilds. When a cache binding or backend stripe size changes, `forced_frontend_updates` rebuilds an otherwise unchanged frontend. If that replacement fails, `last_frontends` retains the old live driver, but lines 960-964 treat the unchanged frontend spec as proof that the desired binding applied and overwrite `last_bindings` with it; a successfully replaced backend similarly advances `last_backends`. On the next apply of the same desired snapshot, `frontend_rebuild_ids` therefore sees neither binding nor geometry drift, skips the frontend replacement, and reports success while the live driver still uses its old backend/cache/stripe geometry. Preserve the old auxiliary state for failed/deferred forced updates and add a retry regression covering a frontend build failure after binding or stripe-size drift.

jveski:
@copilot+claude-sonnet-5 resolve the merge conflicts in this pull request

imiller31 (approved):

PR #388, opened 2026-06-30
[unbounded-system] Integration: single namespace + operator-driven migration

plombardi89:
## Summary

`unbounded-operator` is the future of installing and upgrading Unbounded. Instead of applying a stack of per-component manifests spread across several namespaces, you install one operator. It reconciles every Unbounded component into a single, configurable `unbounded-system` namespace, driven by `Site` custom resources, and it automatically migrates existing clusters off the legacy split-namespace layout - copying the state that cannot be regenerated and reaping the old namespaces once the new workloads are healthy.

This is the umbrella PR that lands the whole `unbounded-system` effort on `main`.

## Why

Unbounded historically installed across several namespaces (`unbounded-kube` for machina/metalman/storage, `unbounded-net` for net, plus orca/inventory/machine-ops/gantry). The split is historical, not meaningful: it fragments RBAC, discovery, NetworkPolicy, and day-2 ops, and leaves no single namespace that represents "Unbounded" on a cluster. This consolidates everything onto one configurable namespace and provides a safe, automated path to move existing clusters onto it (Kubernetes namespaces cannot be renamed in place, so a deliberate migration is required).

## What it does

- **One namespace.** Operator, net, machina, metalman, and storage all run in `unbounded-system`. The namespace is configurable and POD_NAMESPACE-aware end to end (`internal/unbounded.SystemNamespace()`).
- **Operator-driven install/upgrade.** `kubectl unbounded install` deploys the operator; the operator owns CRD lifecycle and installs components at its own version. Components are declared per location via the `Site` API rather than hand-applied manifests.
- **Automatic migration.** Upgrading from the legacy layout, the operator's always-on, leader-elected reaper translates old net-group Sites, copies non-regenerable state, repoints cluster-scoped references, and drains/deletes the legacy namespaces once each component's new workload is Ready.

## Architecture

- **Site API** (`api/machina/v1alpha3/site_types.go`): cluster-scoped `Site` in the machina group carries the networking spec plus `spec.components` (machina / metalman / storage). Networking (unbounded-net) is not a component; it is a cluster singleton the operator runs whenever any Site exists.
- **Operator** (`cmd/unbounded-operator`, `internal/operator`):
  - CRD bootstrap + maintainer: installs/updates CRDs on startup and keeps them Established (`bootstrap.go`).
  - `SiteReconciler`: renders and owns component workloads from embedded manifests into `unbounded-system` (`reconciler.go`).
  - `LegacyReaper`: the migration routine (`migrate.go`).
[...]

PR #464, opened 2026-07-13
[unbounded-system] Operator owns CRD lifecycle and config; harden migration

plombardi89:
## Summary

Follow-up to #386 that was meant to land there. Targets the `feature/unbounded-system` integration branch as part of umbrella #388.

Makes `unbounded-operator` self-sufficient for CRD lifecycle, moves environment-specific settings into a ConfigMap, and hardens reconciliation and legacy migration against rollout, ownership, readiness, and configuration races.

## Operator owns CRDs

- Add `operator.BootstrapCRDs`: at startup the operator server-side applies all 12 CRDs embedded in the machina and net manifests before starting the manager.
- Bound the complete bootstrap operation, including CRD applies and establishment, with a four-minute timeout.
- Add a startup probe whose budget exceeds the complete bootstrap timeout so liveness cannot restart the operator mid-bootstrap.
- `kubectl unbounded install` no longer applies CRDs or exposes `--skip-crds`; the breaking CLI change is documented.
- `install` waits for a complete operator rollout, rejecting the old-ReplicaSet-available/new-pod-failing upgrade case, and then waits for CRD establishment.
- The Site reconcile loop skips CRD objects because CRDs are owned exclusively by startup bootstrap.

## Config via ConfigMap

- Add `unbounded-operator-config`, surfaced to the Deployment through `envFrom`.
- `--api-server-endpoint` and `--reap-legacy-resources` default from `UNBOUNDED_API_SERVER_ENDPOINT` and `UNBOUNDED_REAP_LEGACY_RESOURCES`; explicit flags retain precedence.
- Invalid `UNBOUNDED_REAP_LEGACY_RESOURCES` values now fail startup rather than silently enabling destructive migration.
- Apply the operator ConfigMap before the Deployment to prevent a new pod from starting with stale environment values.
- Stamp the resolved config hash on the operator pod template so endpoint changes trigger a rollout in both `install` and rendered-manifest workflows.
- Restore `UNBOUNDED_OPERATOR_API_SERVER_ENDPOINT` to the Make/render path and make endpoint forwarding and YAML rendering safe for shell- and YAML-sensitive values.

## Machina config propagation

- Merge the configured API endpoint into existing or migrated `machina-config` while preserving all other custom settings and comments.
- Hash the exact stored `config.yaml` and stamp it on the Machina Deployment so configuration changes roll Machina, which reads its subPath-mounted config only at startup.
- Watch `machina-config` changes and enqueue Sites without re-enabling status-only Site reconciliation.
- During migration, apply the configured endpoint to copied legacy Machina config.
- Reap legacy Machina only after the replacement Deployment has the current config hash and a completed rollout.

## Reconciliation reliability

- Preserve per-component reconciliation errors while still publishing all Site status conditions.
- Return joined component and status-patch errors so controller-runtime rate-limits and retries transient failures even with `GenerationChangedPredicate` filtering status-only updates.

## Migration hardening

- Detect legacy per-site Metalman by deterministic Deployment name or canonical/deprecated Site label.
- Preserve the legacy `--dhcp-auto-interface` setting in the translated Site; malformed or conflicting values fail closed.
- Gate Metalman reaping on creation of every corresponding replacement Deployment without requiring readiness, avoiding host-port cutover deadlock.
- Require storage replacements to be current and ready. A zero-desired per-Site DaemonSet is accepted only when no Node InternalIP matches that Site node CIDRs; matching but not-yet-labeled Nodes continue to block reaping.
- Make the net controller update SiteNodeSlice ownership even when slice contents are unchanged.
- Delete the legacy net-group Site CRD only after the new net controller is Available and every existing SiteNodeSlice has the exact current v1alpha3 Site UID/GVK owner reference.
- Warn and emit an Event when a drained legacy namespace still contains non-operator workloads before namespace deletion.
- Add the missing read-only RBAC for StatefulSets, Nodes, and SiteNodeSlices, plus `events.k8s.io` write access for the modern event recorder.
- Reconcile Sites on generation changes only, avoiding full net/machina re-application on status-only updates.
- Retarget embedded namespace references in service-account usernames and flag values, not only exact-value matches.

## Tests

Added or extended coverage for:

- all embedded CRDs, establishment polling, and apply-phase timeout cancellation;
- startup probe budget and operator ConfigMap-before-Deployment ordering;
- complete Deployment rollout detection and endpoint-driven rollout hashes;
- shell/YAML-sensitive endpoint rendering;
- strict environment parsing and explicit flag precedence;
- reconciliation error retries and joined status-patch errors;
- Machina config preservation, endpoint updates, hashes, and rollout gates;
- Metalman detection and DHCP mode preservation;
- zero-node storage readiness and unconverged-node blocking;
- SiteNodeSlice owner migration and legacy CRD deletion gates;
- StatefulSet foreign-workload detection and rendered RBAC.

The kind-based operator reaper e2e now validates the migration against a real Kubernetes API server, including preserved Metalman mode and the Machina config rollout gate.

## Verification

- `make fmt`
- `make build`
- `make lint`
- `make test`
- race tests for all modified Go packages
- `go test -tags=e2e ./e2e/operator/... -run TestOperatorReaperMigration -count=1 -v`

### DIFF
--- a/deploy/unbounded-operator/03-configmap.yaml.tmpl
+++ b/deploy/unbounded-operator/03-configmap.yaml.tmpl
@@ -25,7 +25,7 @@ data:
   # cannot be derived in-cluster. Set this before enabling machina provisioning.
   # Rendered from .APIServerEndpoint so make/GitOps deploys can set it via
   # --set APIServerEndpoint=; `kubectl unbounded install` overrides it in place.
-  UNBOUNDED_API_SERVER_ENDPOINT: "{{ default "" .APIServerEndpoint }}"
+  UNBOUNDED_API_SERVER_ENDPOINT: {{ default "" .APIServerEndpoint | quote }}
   # Whether the operator translates legacy net-group Sites, migrates state into
   # this namespace, and reaps the pre-consolidation namespaces.
   UNBOUNDED_REAP_LEGACY_RESOURCES: "true"
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,6 @@ UNBOUNDED_OPERATOR_BIN=bin/unbounded-operator
 UNBOUNDED_OPERATOR_CMD=./cmd/unbounded-operator
 UNBOUNDED_OPERATOR_IMAGE ?= $(CONTAINER_REGISTRY)/unbounded-operator:$(VERSION)
 UNBOUNDED_OPERATOR_NAMESPACE ?= $(UNBOUNDED_NAMESPACE)
-UNBOUNDED_OPERATOR_API_SERVER_ENDPOINT ?=
 UNBOUNDED_OPERATOR_MANIFEST_TEMPLATES_DIR := deploy/unbounded-operator
 UNBOUNDED_OPERATOR_MANIFEST_RENDERED_DIR  := deploy/unbounded-operator/rendered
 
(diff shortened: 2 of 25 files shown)


deploy/unbounded-operator/03-configmap.yaml.tmpl
```diff
@@ -0,0 +1,31 @@
...
+  labels:
+    app.kubernetes.io/name: unbounded-operator
+    app.kubernetes.io/component: controller
+data:
+  # External Kubernetes API endpoint advertised to provisioned machines. Remote
+  # nodes dial this to join, so it must be reachable from outside the cluster and
+  # cannot be derived in-cluster. Set this before enabling machina provisioning.
+  # Rendered from .APIServerEndpoint so make/GitOps deploys can set it via
+  # --set APIServerEndpoint=; `kubectl unbounded install` overrides it in place.
+  UNBOUNDED_API_SERVER_ENDPOINT: {{ default "" .APIServerEndpoint | quote }}
```
jveski, line 28: We discover this from kube-public in metalman, might be worth aligning on. I'm indifferent as to which approach we take as long as it's consistent.

jveski (approved):

PR #473, opened 2026-07-14
net: harden SiteNodeSlice ownership and reconcile against delete/recreate races

plombardi89:
## Summary
Hardens the net Site controller's SiteNodeSlice ownership and reconcile loop against ownership-admission failures and delete/recreate races.

## Findings addressed
- **Owner ref requires `sites/finalizers` RBAC the controller lacks:** set `blockOwnerDeletion: false` (accept nil/false, repair true) so slice writes are no longer rejected by ownerRef admission.
- **No UID guard -> mutating a same-name replacement Site:** every slice/status/finalizer write now does a live GET + expected-UID check.
- **Unconditioned extra-slice deletion:** deletes now use UID + ResourceVersion preconditions and revalidate the live `siteName`.
- **Full-object Update clobbered foreign metadata:** mutate a deep copy of the live slice.
- **Dirty flag lost on error:** failures propagate and re-dirty so the next pass retries; per-site status/finalizer errors are aggregated.
- **Status under-counted gateway nodes**, `[]string` embedded in unstructured content, missing orphan-slice cleanup, and missing slice-informer wiring - all fixed.

## Verification
`make fmt` / `make lint` / `make build` / `make test` green; `go test -race ./internal/net/controller/...` green.

## Merge notes
Independent, compile-standalone. Runtime-coupled to the operator config/migration PR (reaper owner-ref predicate); recommend merging this **before** that PR. Non-destructive in either order.

### DIFF
--- a/internal/net/controller/site_controller.go
+++ b/internal/net/controller/site_controller.go
@@ -138,9 +138,8 @@ type SiteController struct {
 	// Mutex to prevent concurrent slice updates
 	sliceUpdateLock sync.Mutex
 
-	// slicesDirty is set when syncNode changes node state and cleared when
-	// updateAllSiteSlices runs. The periodic loop checks this to coalesce
-	// slice updates instead of rebuilding after every single node sync.
+	// slicesDirty is set when node or slice state changes. The periodic loop
+	// atomically consumes it to coalesce updates and restores it after failures.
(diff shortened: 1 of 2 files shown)


plombardi89:

Notes:

cchildress (approved):

PR #474, opened 2026-07-14
operator: preserve migrated config, hash full ConfigMap payloads, and harden the legacy reaper

plombardi89:
## Summary
Makes the legacy-migration reaper safe against stale reads and TOCTOU races, preserves migrated `unbounded-net-config`, and moves all config rollout decisions onto a full-payload hash.

### Migration reaper (`migrate.go`)
- Route every destructive read through an **uncached APIReader**; `namespaceExists` now **fails closed** `(bool, error)`.
- `createIfAbsent` Gets into a **freshly zeroed** object and compares the complete payload before treating a target as migrated - closes a silent data-loss path where the real client merges Get results into the desired object's maps. *(Includes the most recent review finding.)*
- `copyConfigMapByName` refuses to leave a dangling cloud-init ref; `translateSite` requires spec-equivalence and is re-verified live before the legacy Site CRD delete.
- machina/net/storage readiness gates re-check ResourceVersion + full-payload hash (TOCTOU); DaemonSet readiness requires `UpdatedNumberScheduled`.
- Foreign-workload audit broadened to pods/RCs/RSs/jobs/cronjobs, fail-closed, skipping controller-owned/terminating legacy descendants (+ RBAC grants).
- Secret `Immutable` copied/compared; fixed the inverted SiteNodeSlice owner-ref predicate that stalled migration.

### Reconciler (`reconciler.go`)
- Preserve a live `unbounded-net-config` instead of force-applying embedded defaults; stamp the config hash on both net workloads.
- Watch all managed ConfigMaps and singleton workloads; reconcile retained net/machina without recreating them from nothing; optimistic-locked config patches.

## Verification
`make fmt` / `make lint` / `make build` / `make test` green; `go test -race` on operator/deploy/cmd-operator green; new regression test `TestCreateIfAbsentDetectsSubsetLiveTarget` fails on the old code, passes on the fix.

## Merge notes
Recommend merging **after** the net SiteNodeSlice PR (owner-ref coupling). Touches `cmd/unbounded-operator/main.go` and `deploy/unbounded-operator/render_test.go` in hunks that may need a trivial rebase against the CRD-lifecycle/install PR.


plombardi89:

Notes:

cchildress (approved):

internal/operator/migrate.go
```diff
@@ -423,32 +441,37 @@ func (r *LegacyReaper) translateSites(ctx context.Context, logger logr.Logger) e
...
 
-	err := r.Get(ctx, client.ObjectKey{Name: name}, existing)
+	err = r.liveReader().Get(ctx, client.ObjectKey{Name: name}, existing)
 	if err == nil {
+		existingNetworkingSpec, err := siteSpecWithoutComponents(existing)
+		if err != nil {
+			return fmt.Errorf("read machina site %s spec: %w", name, err)
+		}
+
+		if !reflect.DeepEqual(networkingSpec, existingNetworkingSpec) {
```
jveski, line 461: You're probably fine with DeepEqual here but consider swapping it out for this in case there are fields with equality logic not defined by the Go type: https://pkg.go.dev/k8s.io/apimachinery/pkg/api/equality

jveski: Oh nvm- it's unstructured anyway.

jveski (approved):

PR #475, opened 2026-07-14
operator: own CRD lifecycle with periodic maintenance and idempotent install repair

plombardi89:
## Summary
Gives the operator ongoing ownership of its CRD set and makes `kubectl unbounded install` correctly repair CRDs and roll config.

## Findings addressed
- **No ongoing CRD repair:** leader-elected `CRDMaintainer` re-applies/re-establishes the full owned CRD set on an interval.
- **Incomplete CRD contract:** exported `RequiredCRDNames` (all 12) is the single source of truth; bootstrap waits on the full set; terminating CRDs are rejected as established (operator + install client).
- **Install waited on only 7 CRDs; config hash covered only the endpoint:** now waits on all 12 and hashes the **full** operator ConfigMap payload so a change to `UNBOUNDED_REAP_LEGACY_RESOURCES` rolls the operator.
- **Reinstall silently re-enabled the reaper:** preserve the live value via `strconv.ParseBool`, fail closed on invalid before applying.
- **No CRD repair against a healthy operator:** idempotent repair token that only rolls when CRDs are unhealthy and the prior rollout is complete.
- **Rolling update could false-succeed:** operator Deployment pinned to `Recreate` with `rollingUpdate: null`; reap flag templated and threaded through the Makefile + render tests.

## Verification
`make fmt` / `make lint` / `make build` / `make test` green; `go test -race` on operator/install/cmd-operator/deploy green.

## Merge notes
Compile-independent of the operator config/migration PR. Touches `cmd/unbounded-operator/main.go` and `deploy/unbounded-operator/render_test.go`; the second of this/the config-migration PR to merge will need a trivial `render_test.go` conflict resolution (keep both PRs' assertions).

### DIFF
--- a/internal/operator/bootstrap.go
+++ b/internal/operator/bootstrap.go
@@ -37,6 +37,13 @@ const (
 	// crdEstablishedTimeout preserves the full establishment window after the
 	// apply phase while CRDBootstrapTimeout bounds the complete operation.
 	crdEstablishedTimeout = 2 * time.Minute
+
+	defaultCRDMaintenanceInterval = time.Minute
+
+	// defaultCRDMaintenanceFailures is the number of consecutive CRD maintenance
+	// failures tolerated before the maintainer stops the manager and lets the
+	// pod restart.
+	defaultCRDMaintenanceFailures = 3
 )
 
 // CRDBootstrapTimeout bounds the complete CRD bootstrap, including manifest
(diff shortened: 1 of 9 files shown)


plombardi89:

Notes:

internal/operator/bootstrap.go
```diff
@@ -62,33 +81,76 @@ func BootstrapCRDs(ctx context.Context, c client.Client) error {
...
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			if err := bootstrap(ctx, m.Client); err != nil {
+				if ctx.Err() != nil {
+					return nil
+				}
+
+				return fmt.Errorf("maintain CRDs: %w", err)
```
jveski: I'd at least retry a few times before resorting to crashing the pod

plombardi89: Good call. Updated `CRDMaintainer` to tolerate transient failures: it now logs and retries on the next interval, and only stops the manager after `MaxFailures` (default 3) consecutive failures, resetting on success. This keeps the pod-restart escape hatch for a persistently stuck/terminating CRD while no longer crashing on a single blip. Added `TestCRDMaintainerRetriesTransientFailureWithoutCrashing` and updated `TestCRDMaintainerReturnsBootstrapError` to assert the 3-strikes behavior.

PR #500, opened 2026-07-16
gantry: cold-start cascade tuning + ACR credential provider fixture

vpatelsj:
Gantry cascade / cold-start:
- Configurable prefetch puller replication (prefetch_puller_replicas, default 8) so a cold thundering-herd fans out from N seeds instead of bottlenecking on one.
- Peer re-discovery loop (peer_rediscover_budget/backoff) plus a transfer serve cap (transfer_max_concurrent_serves, default 100) with 429 + Retry-After load-shedding to drive the swarm cascade.
- Shuffle peer providers so fetches spread across seeds; fix self-puller race and loopback peer dial.
- Ship the validated cascade defaults in config.NewDefault() and deploy/gantry/configmap.yaml.

ACR kubelet credential provider fixture (hack/acr-credential-provider):
- Provider binary, installer image, and DaemonSet for private ACR pulls without imagePullSecrets, plus Makefile build/image targets.

Playbooks:
- 300-node AKS cluster playbook and standalone Gantry install playbook.


jwilder (dismissed):

hack/acr-credential-provider/internal/acrcredentialprovider/azure.go
```diff
@@ -0,0 +1,118 @@
...
+
+type AzureCredentialSource struct {
+	TokenCredential azcore.TokenCredential
+	HTTPClient      *http.Client
+	AADScope        string
+	TenantID        string
+}
+
+func NewDefaultAzureCredentialSource() (*AzureCredentialSource, error) {
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
```
bcho: this will not actually work for env without Azure IMDS as you need to pass in the service account token in those env... i prefer to include this https://github.com/kubernetes-sigs/cloud-provider-azure/tree/master/cmd/acr-credential-provider or build on on top of it instead so we will have full coverage of different scenarios like https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/4412-projected-service-account-tokens-for-kubelet-image-credential-providers/README.md

vpatelsj:
@bcho good call - the bundled provider used `DefaultAzureCredential` (IMDS), so it wouldn't cover non-IMDS environments where you have to pass the projected service-account token (KEP-4412).

I've dropped the custom fixture entirely in ad24b681. On AKS the upstream `cloud-provider-azure` `acr-credential-provider` binary and its `CredentialProviderConfig` are already installed and wired into kubelet on every node, so we now rely on that exact upstream binary instead of shipping our own - which gives us the upstream's full scenario coverage (managed identity, workload identity, projected SA tokens) rather than IMDS-only.

Since we don't need private ACR pulls on generic non-AKS clusters for this work, I removed the provider binary, installer image/DaemonSet, and the Makefile build/image targets, and rewrote the standalone Gantry playbook to rely on the AKS built-in provider (grant the kubelet identity `AcrPull`, no install step). If we later need non-AKS coverage, we'd install or build on top of the upstream provider you linked rather than maintain our own.

bcho (approved):

PR #507, opened 2026-07-16
machineops: add provider-owned Machine references

imiller31:
## Summary

- model Machine host ownership as mutually exclusive `spec.host.netboot`, `spec.host.azure`, or `spec.host.external` variants
- inline the built-in Azure VM resource ID instead of creating a one-field AzureMachine CRD
- keep `host.external.machineRef` as an optional escape hatch for providers with meaningful provider-owned schema/status/reconciliation
- resolve canonical and released legacy host shapes through one controller resolver and one Metalman netboot accessor
- snapshot external machine reference identity and resolved host image before provider execution
- support Azure and OCI power/replacement operations, resumable custom providers, credential selection, conflict serialization, and replacement bootstrap

## Compatibility

The released top-level `spec.pxe`, `spec.provider`, and `spec.providerID` fields remain readable fallbacks. New writes use `spec.host`; canonical host ownership cannot be mixed with those legacy fields. Migration tooling is intentionally deferred to a separate change.

The selected host variant is immutable. Azure `resourceID`, external `provider`, and external `machineRef` are immutable; external `providerID` remains mutable so replacement providers such as OCI can hand off to a new resource identity.

MachineConfiguration and MachineConfigurationVersion use a distinct provider-neutral host configuration containing only the image. Per-machine host ownership never enters the versioned configuration template.

## Azure end-to-end validation

Ran the real machine-ops controller against a disposable Kind cluster and disposable Azure `Standard_B2s` VM. Verified successful completion and Azure state for:

- `HostPowerOff` (`VM stopped`)
- `HostPowerOn` (`VM running`)
- `HostReboot` (`VM running`)
- `HostReplace` (`Running/Succeeded`)

Replacement retained the VM ARM resource ID and NIC, created a new VM identity and OS disk, and applied the requested Marketplace image. All disposable Azure and Kind resources were removed afterward. The cloud operation implementation is unchanged by the later host-union storage revision; final union routing and validation were separately exercised against a real Kubernetes API server.

The live replacement test exposed and fixed two payload bugs:

1. Azure SDK models shared nested pointers, so preparing the replacement create payload could mutate `osDisk.name` in the earlier pre-delete update payload. Delete and create payloads are now independently cloned, with a regression test.
2. The replacement create payload retained the deleted managed disk ID while using `CreateOption=FromImage`. The ID is now cleared while allowed managed-disk settings remain preserved.

## Validation

- real Azure end-to-end test covering power off/on, reboot, and destructive replacement
- real Kind/operator bootstrap smoke: 12 CRDs established and no AzureMachine CRD
- Kubernetes schema smoke accepted Azure, external, and netboot variants
- schema smoke rejected mixed owners and immutable identity changes while allowing external providerID handoff
- full repository `go test -race ./...`
- full repository golangci-lint (`0 issues`)
- affected API generation and rendered-manifest verification
- repository-wide compile and `git diff --check`

The existing Azure clone/delete/create lost-response crash window is intentionally unchanged by this PR.


jveski:
I like the idea here - just a few concerns about field sprawl in the machine type. Ideally we could use the same provider model for metalman/machina machines without the overhead of a second CR.

What do you think about using kind of `oneOf` semantics to do something like this?

```yaml
kind: Machine
# ...
spec:
  kubernetes: {}
  agent: {}
  host:
     netboot: # or something else to represent metalman
       image: ghcr.io/azure/host-ubuntu24:latest 
       redfish:
         host: 10.10.10.10

---

kind: Machine
# ...
spec:
  kubernetes: {}
  agent: {}
  host:
     azure:
       resourceID: ...

---

kind: Machine
# ...
spec:
  kubernetes: {}
  agent: {}
  host:
     external:
       providerID: ""
       machineRef:
         kind: ...
```

jveski (approved):

PR #513, opened 2026-07-20
Preconfigure Gantry containerd hosts in agent

vpatelsj:
## Summary
- preconfigure agent-provisioned nodes with a managed containerd `_default/hosts.toml` that points at Gantry on `127.0.0.1:5000`
- add `AgentConfig.Gantry.Disabled` as a breakglass flag to skip writing the Gantry hosts file
- refuse to overwrite unmanaged `_default/hosts.toml` while leaving per-registry hosts config untouched
- document that `deploy/gantry/node-config.yaml` remains for standalone/non-agent-managed nodes

## Validation
- `GOTOOLCHAIN=auto go test ./pkg/agent/config ./internal/provision ./pkg/agent/goalstates ./pkg/agent/phases/nodestart ./cmd/agent/internal/daemon`
- `GOTOOLCHAIN=auto make lint`
- `git diff --check`
- manual absent-mirror validation on temporary 1-node AKS cluster `ubg07201348`:
  - node: `aks-nodepool1-24892595-vmss000000`
  - containerd: `2.3.2-1`
  - installed the proposed `_default/hosts.toml` while Gantry was not installed/listening
  - pulled `docker.io/library/busybox:latest` with host `ctr --namespace k8s.io`
  - observed `ABSENT_MIRROR_TEST_PASS`
  - restore trap reported `removed temporary hosts.toml`

## Cleanup
- deleted the temporary test pod
- deleted temporary AKS resource groups `rg-ub-gantry-07201348` and `rg-ub-gantry-nodes-07201348`

### DIFF
--- a/pkg/agent/phases/nodestart/cri.go
+++ b/pkg/agent/phases/nodestart/cri.go
@@ -26,7 +26,15 @@ var assets embed.FS
 
 var assetsTemplate = template.Must(template.New("assets").ParseFS(assets, "assets/*"))
 
-const nvidiaRuntimeDropInName = "99-nvidia-runtime.toml"
+const (
+	nvidiaRuntimeDropInName  = "99-nvidia-runtime.toml"
+	gantryHostsManagedMarker = "# Managed by unbounded-agent for Gantry."
+	gantryHostsConfig        = gantryHostsManagedMarker + `
+[host."http://127.0.0.1:5000"]
+  capabilities = ["pull", "resolve"]
+  dial_timeout = "200ms"
+`
+)
 
 type configureContainerd struct {
 	goalState *goalstates.NodeStart
@@ -44,6 +52,12 @@ func (c *configureContainerd) Name() string { return "configure-containerd" }
 func (c *configureContainerd) Do(_ context.Context) error {
 	if err := c.ensureContainerdConfig(); err != nil {
 		return fmt.Errorf("ensure containerd config: %w", err)
+	}
+
+	if !c.goalState.Gantry.Disabled {
+		if err := c.ensureGantryHostsConfig(); err != nil {
+			return fmt.Errorf("ensure Gantry containerd hosts config: %w", err)
+		}
 	}
 
 	if err := c.ensureContainerdServiceUnit(); err != nil {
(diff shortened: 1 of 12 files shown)


pkg/agent/phases/nodestart/cri.go
```diff
@@ -26,7 +26,15 @@ var assets embed.FS
 
 var assetsTemplate = template.Must(template.New("assets").ParseFS(assets, "assets/*"))
 
-const nvidiaRuntimeDropInName = "99-nvidia-runtime.toml"
+const (
+	nvidiaRuntimeDropInName  = "99-nvidia-runtime.toml"
+	gantryHostsManagedMarker = "# Managed by unbounded-agent for Gantry."
+	gantryHostsConfig        = gantryHostsManagedMarker + `
+[host."http://127.0.0.1:5000"]
```
jveski, line 33: Would prefer a UDS listener for the sake of security and to avoid port conflicts with customer workloads. Doesn't necessarily need to land as part of this change though

jveski, line 33: If gantry is unavailable or for some reason not installed, how does the fallback behave?

e.g. does it wait for a timeout before falling through to the next host? And is that timeout configurable? I'm thinking about cases where waiting for that fallback might cause the overall container pull operation to time out.

bcho: +1 can we avoid using a generic port here

pkg/agent/phases/nodestart/cri.go
```diff
@@ -78,6 +90,37 @@ func (c *configureContainerd) ensureContainerdConfig() error {
 	return utilio.WriteFile(dest, buf.Bytes(), 0o644)
 }
 
+func (c *configureContainerd) ensureGantryHostsConfig() error {
+	dest := filepath.Join(c.goalState.MachineDir, goalstates.ContainerdDefaultHostsPath)
```
bcho, line 96: can we still add a disable flag to disable this for breakglass case?  For example

```javascript
{
  "Kubelet": {}, // existing kubelet 
  "Gantry": {
    "disabled": bool // default to false
  }
}
```

pkg/agent/phases/nodestart/cri.go
vpatelsj: Agree UDS would be better from a security/port-conflict perspective. I would keep that out of this PR, though: Gantry already exposes the containerd mirror through the existing loopback hostPort contract, and this PR only preconfigures agent nodes to match that contract. Moving to UDS needs a broader Gantry change across listener config, the DaemonSet/hostPath mounts, containerd hosts.toml support validation, docs, and e2e coverage. I think that should be tracked as a follow-up to change Gantry’s endpoint contract rather than blocking this bootstrap prep.

vpatelsj: The file sets `dial_timeout = "200ms"`, so an absent listener should fail fast. I validated the exact absent-Gantry case manually on a temporary 1-node AKS cluster using containerd 2.3.2-1: installed this `_default/hosts.toml`, left Gantry uninstalled/not listening, then ran `ctr --namespace k8s.io images pull docker.io/library/busybox:latest` on the host. The pull succeeded and the test logged `ABSENT_MIRROR_TEST_PASS`; the restore trap then removed the temporary hosts file. I added those details to the PR body.

pkg/agent/phases/nodestart/cri.go
vpatelsj: Added this as `AgentConfig.Gantry.Disabled`. When set, node-start skips writing the Gantry `_default/hosts.toml` entirely, including the unmanaged-file refusal path. I also threaded it through goal-state resolution, AgentConfig deep-copy, and daemon drift detection so changing the breakglass flag after bootstrap is reconciled. Added tests for JSON round-trip/deep-copy, goal-state propagation, disabled skip behavior, and drift.

bcho (approved):

PR #552, opened 2026-07-28
design: add nspawn LocalDNS proposal

bcho:
## Summary

- propose running a two-listener CoreDNS cache inside the active nspawn machine
- define machine resolver, kubelet DNS, host network, NOTRACK, watchdog, resource, and lifecycle behavior based on AgentBaker LocalDNS
- define online/offline CoreDNS artifact acquisition, plugin verification, metrics, preflight, and AKS Flex Node ownership boundaries
- document remaining AgentBaker parity gaps and intentional architectural differences

## Validation

- `git diff --check`
- documentation-only change


designs/agent-nspawn-localdns.md
```diff
@@ -0,0 +1,1218 @@
...
+
+    cache 30 {
+        success 9984
+        denial 9984
+        serve_stale 3600s verify
+        servfail 0
+    }
+
+    loop
+    prometheus {{ .MetricsAddress }}
```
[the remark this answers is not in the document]
phealy: Agreed.

designs/agent-nspawn-localdns.md
```diff
@@ -0,0 +1,1218 @@
...
+addresses. `OUTPUT` covers node services and host-network pods;
+`PREROUTING` covers regular pod traffic. The complete desired set contains eight
+rules:
+
+```text
+2 chains x 2 listener addresses x 2 protocols = 8 rules
+```
+
+Each rule carries the comment `unbounded-localdns: skip conntrack`. The host
+network oneshot reconciles the exact desired set using the iptables-compatible
```
phealy: I suggest using nftables natively if available instead of operating via iptables-compat.

designs/agent-nspawn-localdns.md
```diff
@@ -0,0 +1,1218 @@
...
+
+The native CoreDNS endpoint provides the DNS, cache, forwarding, process, and
+runtime metrics needed for the initial feature. A second exporter and port are
+not justified without a compatibility requirement for AgentBaker-specific
+metric names.
+
+## Open question
+
+- Should node upstream discovery use a local stub such as `127.0.0.53` when
+  present, or require resolver-backend-specific discovery of direct upstreams?
```
phealy: Don't send traffic via systemd-resolved; multiple layers of caching are too hard to reason about.

phealy (dismissed):
Basic design sounds like what aks-local-dns does but adapted quite well for unbounded. A few minor details commented.

designs/agent-nspawn-localdns.md
```diff
@@ -0,0 +1,1218 @@
...
+type AgentLocalDNSConfig struct {
+    Enabled           bool   `json:"Enabled"`
+    NodeListenerIP    string `json:"NodeListenerIP,omitempty"`
+    ClusterListenerIP string `json:"ClusterListenerIP,omitempty"`
+    MetricsAddress       string   `json:"MetricsAddress,omitempty"`
+    CPULimitInMilliCores *int     `json:"CPULimitInMilliCores,omitempty"`
+    MemoryLimitInMB      *int     `json:"MemoryLimitInMB,omitempty"`
+    RequiredPlugins      []string `json:"RequiredPlugins,omitempty"`
+    CorefileTemplate     string   `json:"CorefileTemplate,omitempty"`
+}
```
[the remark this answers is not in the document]
bcho: Fixed the example to use gofmt-style field and type alignment.

designs/agent-nspawn-localdns.md
bcho: Updated the design to prefer native nftables with a dedicated agent-owned table and atomic reconciliation. The iptables-compatible frontend is retained only as a fallback when the native frontend is unavailable.

designs/agent-nspawn-localdns.md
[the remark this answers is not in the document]
bcho: Removed the Prometheus directive from the cluster-listener server block. The design now clarifies that the single process-wide endpoint exports metrics for both CoreDNS server blocks.

designs/agent-nspawn-localdns.md
bcho: Updated upstream discovery to require direct upstream addresses. The initial scope now supports direct nameservers in /etc/resolv.conf and the systemd-resolved upstream file when there is no split-DNS policy. Local caching stubs, unsupported resolver layouts, and split-DNS configurations fail preflight rather than being flattened or used as forwarding destinations.

vpatelsj (approved):

PR #567, opened 2026-08-03
docs: add control plane upgrade workflow design doc

plombardi89:
## Summary

Adds `designs/upgrade-workflow.md`, documenting how the Unbounded control plane is upgraded: the `unbounded-operator` and the five components it manages (`net`, `machina`, `gantry`, `metalman`, `storage`), plus the operator-owned CRDs.

This is **descriptive of current behavior, not a proposal**. Every `file:line` citation was verified against the source. Scope is deliberately limited to the operator and what it manages; the host `unbounded-agent` and node repave are separate update planes and are called out as out of scope.

Covers: the version model (operator version is the fleet version), image resolution, embedded manifests, SSA with `ForceOwnership`, CRD bootstrap ordering, config preservation and config-hash-driven rollouts, per-workload rollout strategies and their availability implications, verification, and six design constraints.

## Review checklist

Three items I'd like the team to weigh in on. Each states current behavior with citations; none has a recommendation attached.

- [ ] **Should `kubectl unbounded install` be renamed to `kubectl unbounded apply`?**
  The command is already idempotent and is already the upgrade path: `nightly.yaml` computes `MODE=init|upgrade` and runs `install` for both (`.github/workflows/nightly.yaml:490`), as does `release-upgrade.yaml:361`. It preserves existing config across re-runs (`cmd/kubectl-unbounded/app/install.go:284-338`), and it no longer installs CRDs since the operator owns them (`install.go:140-143`). Against: it takes no `-f` and does not behave like `kubectl apply`, so the analogy is partial. Cost is breadth - two workflows, `hack/operator-upgrade-e2e/e2e.py:611`, and six docs pages. Cobra `Aliases` offers a non-breaking path; the command is registered at `cmd/kubectl-unbounded/app/cmd.go:29` and declares no aliases today.

- [ ] **Do we need a mechanism to preserve user customizations of managed workloads (e.g. custom requests/limits)?**
  Today there is no way to do this. Every managed manifest sets `resources` explicitly (`net/controller:123`, `net/node:100`, `machina:57`, `gantry:141` and `:222`, `storage-supervisor:93`), and the operator applies the full object with `ForceOwnership` (`internal/operator/component/env.go:232-239`), so `resources` is operator-owned. Reversion is near-immediate: `ManagedWorkloadPredicate` fires on generation change (`internal/operator/component/watch.go:129-140`), so a `kubectl edit` enqueues the reconcile that undoes it. No pause, opt-out, or unmanaged annotation exists in `internal/operator/`.
  Two precedents already exist but are not generalized: ConfigMaps are create-only and preserved (`ensureConfig`), and `metalman.replicas` is a typed workload-shape knob on the `Site` API. Relevant SSA detail: a field **omitted** from the applied configuration is not removed if another field manager owns it, so dropping `resources` from the embedded manifests would make user-set values stick.

- [ ] **How does a user set the registry override when installing or upgrading via plain `kubectl apply`?**
  Currently two separate edits to `unbounded-operator-<TAG>.yaml`: `UNBOUNDED_IMAGE_REGISTRY` in the ConfigMap, and the operator's own `image:`. There is no single override. Edited after applying, `envFrom` is read once at pod start, so it also needs `kubectl rollout restart deployment/unbounded-operator`.
  Related: `UNBOUNDED_OPERATOR_IMAGE_REGISTRY ?= ghcr.io` (`Makefile:88`) does not default from `CONTAINER_REGISTRY ?= ghcr.io/azure` (`Makefile:13`), and the release workflow passes only `CONTAINER_REGISTRY` (`.github/workflows/release.yaml:805-809`). With `Config.Image` hardcoding `/azure/` (`internal/operator/component/env.go:74-76`), a fork publishing to `ghcr.io/myorg` gets an operator at `ghcr.io/myorg/unbounded-operator:<TAG>` resolving components to `ghcr.io/azure/<component>:<TAG>`. This works today only because the repo owner is `azure`. There is no Helm or Kustomize for Unbounded, so GitOps users have no standard patch mechanism.

## Notes

Documentation only - no code, manifest, or workflow changes.

### DIFF
--- a/designs/upgrade-workflow.md
+++ b/designs/upgrade-workflow.md
@@ -1,0 +1,362 @@
+# Control plane upgrade workflow
+
+Status: current behavior as implemented
+
+This document describes how the Unbounded control plane is upgraded: the
+`unbounded-operator` and the components it manages. It covers both the
+high-level model and the mechanics behind it.
+
+## Summary
+
+The upgrade unit is the operator.
+
+The operator's compiled-in version is the version of every component it manages.
+Upgrading a cluster means replacing one image: the `unbounded-operator`
+Deployment. The operator then re-applies every component workload at its own
+version, and Kubernetes performs the rollouts.
+
+Two consequences follow from that and are worth stating up front:
+
+- **The operator does not upgrade itself.** Something external replaces its
+  image. That is `kubectl unbounded install` run from a newer plugin, or a
+  direct apply of the released operator manifest. There is no
+  `kubectl unbounded upgrade` command.
+- **There is no per-component version selection.** The `Site` API exposes no
+  version, image, tag, or channel field. Components are upgraded as a set, in
+  lockstep with the operator.
+
+## Scope
+
+The operator owns two things: the CRDs, and five component workloads.
+
+| Component | Kind | Workloads |
+|-----------|------|-----------|
+| `net` | cluster singleton | `Deployment/unbounded-net-controller`, `DaemonSet/unbounded-net-node` |
+| `machina` | cluster singleton | `Deployment/machina-controller` |
+| `gantry` | cluster singleton | `DaemonSet/gantry`, `DaemonSet/gantry-containerd-config` |
+| `metalman` | per-Site | `Deployment/metalman-controller-<site>` |
+| `storage` | per-Site | `DaemonSet/unbounded-storage-supervisor-<site>` |
+
+Registered in `operator.DefaultRegistry` (`internal/operator/reconciler.go:66-78`).
+Cluster components reconcile from the full set of `Site` objects on every pass;
+per-Site components reconcile once per `Site`.
+
+The operator also owns twelve CRDs, six in the `unbounded-cloud.io` group and six
+in `net.unbounded-cloud.io` (`internal/operator/bootstrap.go:49-62`).
+
+**Out of scope for this document.** The host-resident `unbounded-agent` binary
+and node OS/kubelet changes are *not* part of this model. They are separate
+update planes with their own triggers (`MachineOperation` of kind `AgentUpgrade`,
+and `MachineConfiguration` repave respectively) and are not driven by an operator
+upgrade. See `designs/agent-upgrade.md` and
+`designs/lifecycle-operations.md`. Components deployed outside the operator
+(`machine-ops-controller`, inventory, orca, playpen) are also out of scope.
+
+## High-level workflow
+
+```
+  release tag
+      |
+      +--> container images pushed as <registry>/azure/<component>:<TAG>
+      +--> every Go binary stamped with internal/version.Version = <TAG>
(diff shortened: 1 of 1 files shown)


designs/upgrade-workflow.md
```diff
@@ -0,0 +1,362 @@
...
+| `storage` | per-Site | `DaemonSet/unbounded-storage-supervisor-<site>` |
+
+Registered in `operator.DefaultRegistry` (`internal/operator/reconciler.go:66-78`).
+Cluster components reconcile from the full set of `Site` objects on every pass;
+per-Site components reconcile once per `Site`.
+
+The operator also owns twelve CRDs, six in the `unbounded-cloud.io` group and six
+in `net.unbounded-cloud.io` (`internal/operator/bootstrap.go:49-62`).
+
+**Out of scope for this document.** The host-resident `unbounded-agent` binary
```
jwilder, line 47: Should this be in scope?  Could the operator use the underlying MachineOperations to orchestrate the agent upgrade?

designs/upgrade-workflow.md
```diff
@@ -0,0 +1,362 @@
...
+The registry is separate and resolves from `--image-registry`, then
+`$UNBOUNDED_IMAGE_REGISTRY`, then `ghcr.io` (`cmd/unbounded-operator/main.go:76`).
+
+### Image resolution
+
+`internal/operator/component/env.go:74-76`:
+
+```go
+func (c Config) Image(repository string) string {
+    return strings.TrimRight(c.ImageRegistry, "/") + "/azure/" + repository + ":" + c.ImageTag
```
jwilder, line 121: Is `/azure/` just an requirement because unbounded is in the Azure org currently?

designs/upgrade-workflow.md
```diff
@@ -0,0 +1,362 @@
...
+The consequence for upgrades is that a retained component continues to be
+version-matched to the operator even with no `Site` enabling it. Removal
+requires an explicit uninstall flow, which does not exist yet.
+
+## Rollout behavior per workload
+
+Once the operator applies the new images, Kubernetes drives each rollout using
+that workload's own strategy. These differ meaningfully.
+
+| Workload | Strategy | Availability during upgrade |
```
jwilder, line 238: I think these may need to be tunable per cluster.  For example, with huge clusters, running a daemonset upgrade one pod at a time may not complete in a reasonable amount of time so might need to bump up maxUnavailable/maxSure.  Some folks might want pit stop vs rolling upgrades too to minimize impact while updates apply.

designs/upgrade-workflow.md
jwilder: I think it's conceptually simpler for a user to just have manage upgrades from the operator interface/config.  Not sure how feasible that is for agent though.

PR #592, opened 2026-08-10
Agent: add host-driven binary upgrade

bcho:
## Summary
- add a hidden `unbounded-agent agent-upgrade [--preflight]` host-driven activation command
- expose reusable, lock-coordinated binary activation APIs for node-side consumers
- coordinate direct and MachineOperation upgrades with the shared lock and pending signal
- add unit coverage and a QEMU/Kind host-upgrade E2E scenario

## Validation
- `go test ./pkg/agent/agentbinary ./pkg/agent/goalstates ./cmd/agent/internal/daemon ./cmd/agent/internal/cmd`
- focused `golangci-lint` for changed Go packages
- `go build -o bin/unbounded-agent ./cmd/agent`
- Python syntax and E2E command registration checks

### DIFF
--- a/cmd/agent/internal/cmd/agentupgrade.go
+++ b/cmd/agent/internal/cmd/agentupgrade.go
@@ -5,10 +5,12 @@ package cmd
 
 import (
 	"context"
+	_ "embed"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
+	"text/template"
 
 	"github.com/spf13/cobra"
 
@@ -18,6 +20,13 @@ import (
 )
 
 const hostAgentBinaryMode = 0o755
+
+//go:embed assets/agent-upgrade-plan.txt.tmpl
+var hostAgentUpgradePlanTemplateText string
+
+var hostAgentUpgradePlanTemplate = template.Must(
+	template.New("host-agent-upgrade-plan").Parse(hostAgentUpgradePlanTemplateText),
+)
 
 type hostAgentUpgradeHandler struct {
 	cmdCtx       *CommandContext
@@ -42,9 +51,12 @@ func newCmdHostAgentUpgrade(cmdCtx *CommandContext) *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:    "agent-upgrade",
-		Short:  "Activate this executable as the host agent daemon",
-		Long:   "Activate this executable as the host agent daemon without creating a MachineOperation.",
+		Use:   "agent-upgrade",
+		Short: "Activate this executable as the host agent daemon",
+		Long:  "Activate this executable as the host agent daemon without creating a MachineOperation.",
+		// Host delivery automation invokes this command from a separately staged
+		// candidate. Keep it out of normal user-facing help because the supported
+		// interactive upgrade surface is the coordinated MachineOperation command.
 		Hidden: true,
 		Args:   cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, _ []string) error {
@@ -85,7 +97,7 @@ func (h *hostAgentUpgradeHandler) execute(ctx context.Context) error {
 			LastGoodPath: paths.LastGoodPath,
 		},
 		CandidatePath: candidatePath,
-		Mode:          hostAgentBinaryMode,
+		BinaryMode:    hostAgentBinaryMode,
 		LockPath:      goalstates.DaemonAgentUpgradeLockPath,
 	}
 	service := h.newService(paths)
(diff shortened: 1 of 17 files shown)


pkg/agent/agentbinary/activation.go
```diff
@@ -0,0 +1,489 @@
...
+			return fmt.Errorf("update current agent symlink: %w", err)
+		}
+	}
+
+	if err := utilio.UpdateSymlink(layout.BinaryPath, layout.CurrentPath); err != nil {
+		return fmt.Errorf("update compatibility agent symlink: %w", err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
bcho: Fixed in 6442a3cd. Activation now snapshots the prior last-good symlink and restores it together with current when a later link switch fails, except when last-good had to be moved to protect a slot being overwritten. BinaryPath is now updated only when configured. Added regression tests for both link restoration and an empty optional BinaryPath.

pkg/agent/agentbinary/activation.go
```diff
@@ -0,0 +1,456 @@
...
+		return nil, fmt.Errorf("create agent activation lock directory: %w", err)
+	}
+
+	file, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR, 0o600)
+	if err != nil {
+		return nil, fmt.Errorf("open agent activation lock: %w", err)
+	}
+
+	if err := unix.Flock(int(file.Fd()), unix.LOCK_EX|unix.LOCK_NB); err != nil {
+		file.Close() //nolint:errcheck // preserve lock acquisition error
```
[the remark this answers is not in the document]
bcho: Fixed in 6442a3cd. The flock failure path now captures file.Close(), joins it with the flock error, and retains ErrActivationInProgress matching for lock contention.

cmd/agent/internal/cmd/agentupgrade.go
```diff
@@ -4,11 +4,172 @@
...
+	}
+	handler.newService = func(paths goalstates.AgentUpgradePaths) agentbinary.DaemonService {
+		return daemon.NewHostDaemonActivationService(handler.cmdCtx.Logger, paths)
+	}
+
+	cmd := &cobra.Command{
+		Use:    "agent-upgrade",
+		Short:  "Activate this executable as the host agent daemon",
+		Long:   "Activate this executable as the host agent daemon without creating a MachineOperation.",
+		Hidden: true,
```
bcho, line 60: add a comment to describe why this command is intended to be hidden

cmd/agent/internal/cmd/agentupgrade.go
```diff
@@ -4,11 +4,172 @@
...
+	path, err := filepath.Abs(path)
+	if err != nil {
+		return "", fmt.Errorf("resolve absolute candidate executable path: %w", err)
+	}
+
+	return filepath.Clean(path), nil
+}
+
+func writeHostAgentUpgradePlan(w io.Writer, plan agentbinary.ActivationPlan) error {
+	if _, err := fmt.Fprintln(w, "Agent upgrade mode: host-driven"); err != nil {
```
bcho: convert to use go template + asset file

hack/agent/e2e-kind/e2e.py
```diff
@@ -3849,6 +3853,107 @@ def validate_node_reboot_operation() -> None:
...
+        f"sudo rm -f {DAEMON_BINARY}; "
+        f"sudo install -m 0755 {legacy_copy} {DAEMON_BINARY}; "
+        f"sudo rm -f {DAEMON_BINARY_CURRENT} {DAEMON_BINARY_LAST_GOOD} "
+        f"{DAEMON_BINARY_BLUE} {DAEMON_BINARY_GREEN}"
+    )
+    legacy_digest = ssh_capture(f"sudo sha256sum {DAEMON_BINARY} | awk '{{print $1}}'").strip()
+
+    log("Running host-driven agent upgrade preflight...")
+    preflight = ssh_capture(f"sudo {remote_candidate} agent-upgrade --preflight")
+    for expected in (
```
bcho: I am fine with not checking the output here, just show the preflight output

pkg/agent/agentbinary/activation.go
```diff
@@ -0,0 +1,549 @@
...
+
+// ErrActivationInProgress indicates that another host agent activation owns
+// the shared activation lock.
+var ErrActivationInProgress = errors.New("host agent activation is already in progress")
+
+// ActivationOptions configures a host-driven daemon binary activation.
+type ActivationOptions struct {
+	Layout        Layout
+	CandidatePath string
+	Mode          os.FileMode
```
bcho: this `Mode` is ambiguous

pkg/agent/agentbinary/activation.go
```diff
@@ -0,0 +1,549 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package agentbinary
```
bcho, line 4: prefer to split this file into two for preflight and actual activation

hack/agent/e2e-kind/e2e.py
bcho: Fixed in 4ad18415. The E2E now prints the preflight output without asserting on its presentation text; it retains the filesystem non-mutation checks.

cmd/agent/internal/cmd/agentupgrade.go
bcho: Fixed in 4ad18415. Added an inline comment explaining that the hidden command is invoked by host delivery automation from a staged candidate, while the normal user-facing path is the coordinated MachineOperation command.

pkg/agent/agentbinary/activation.go
bcho: Fixed in 4ad18415. Renamed ActivationOptions.Mode to BinaryMode and updated the design wording to identify it as the installed binary file mode.

pkg/agent/agentbinary/activation.go
bcho: Fixed in 4ad18415. Split non-mutating planning and its helpers into activation_preflight.go; activation.go now contains the applying transaction, rollback, and shared lock.

cmd/agent/internal/cmd/agentupgrade.go
bcho: Fixed in 4ad18415. The preflight plan is now rendered with text/template from the embedded cmd/agent/internal/cmd/assets/agent-upgrade-plan.txt.tmpl asset.

pkg/agent/agentbinary/activation.go
```diff
@@ -0,0 +1,304 @@
...
+		}
+
+		lastGoodProtected = protected
+		if protected {
+			if err := utilio.UpdateSymlink(opts.Layout.LastGoodPath, plan.RollbackPath); err != nil {
+				return ActivationResult{}, fmt.Errorf("protect active agent as last-good: %w", err)
+			}
+		}
+
+		if err := installFromFile(plan.CandidatePath, plan.TargetPath, mode); err != nil {
```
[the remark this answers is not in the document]
bcho: Fixed in af34807c. ActivateHostDaemon now opens the supplied candidate once under the activation lock, validates that inode, copies it into a private root-owned snapshot directory, and performs preflight, version verification, comparison, and installation from that pinned snapshot. The original staging pathname is not reopened after the snapshot.

jwilder (approved):

PR #603, opened 2026-08-13
net: register admission and the aggregated API only against a live backend

plombardi89:
## What

The net controller exposes three integrations to the Kubernetes API server:

- A validating webhook
- A mutating webhook
- An aggregated `APIService`

These registrations tell the API server to send requests to the `unbounded-net-controller` Service over HTTPS.

This PR ensures those registrations are only created or updated once a controller is actually serving and the API server has the correct CA certificate for it.

## The Problem

Ordering is not readiness.

`tierActivation` already runs the registrations after the Deployment they point at, but "after" only means the apply call returned. A Deployment object accepted by the API server has no pods yet.

The registrations could therefore become active while:

- The controller Pod did not exist yet.
- A new controller revision was still rolling out.
- No controller had acquired leadership.
- The controller's informers and CIDR allocator were not initialized.
- The Service had no working endpoint.
- The registrations had an empty or stale `caBundle`.

The resulting failures were easy to miss.

The webhooks use `failurePolicy: Ignore`, so Kubernetes silently skipped their enforcement when the backend was unavailable. An unavailable `APIService` caused the aggregated API to advertise an API that it could not serve.

The controller also injected its CA into existing registrations only once during startup. That depended on the registrations already existing. If the operator waited to create them until after the controller started, they could remain without a CA indefinitely.

## The Solution

The gate is expressed as **what the net component plans to write**, so it fits the plan-and-execute contract introduced by #587 without the executor needing to know anything about it.

`Plan` reads the live backend before it decodes any manifest, and the answer changes what the pass emits:

1. **Read.** `readBackendState` asks whether the controller can serve the traffic its registrations would send it.
2. **Decode.** `applyMutator` drops the two webhook configurations and the `APIService` while the backend is not serving, and stamps the published CA into them when it is.
3. **Report.** `registrationVerdict` returns the planning verdict: `Reconciled` when serving, `BackendNotReady` when a registration is missing or unusable, and a poll interval either way while withholding.

This is the same shape as `planConfig`, which the component already used: a live read whose answer changes what the pass emits, and whose value is stamped into objects the pass will apply.

```text
Read the live backend
          |
          v
Decode manifests, withholding or stamping the registrations
          |
          v
Emit one plan; the executor writes it in tier order
```

Because withholding is a property of the plan rather than of a second apply, it is visible in `Plan.Summary()` and has a golden test of its own.

### What readBackendState checks

- The controller has published its serving CA.
- Kubernetes has observed the current Deployment generation.
- The current Deployment revision is fully rolled out and Ready.
- The controller Service has the expected HTTPS endpoint.
- The endpoint targets a live, Ready, non-terminating Pod.
- The Pod belongs to the current controller Deployment.

### On reading before the pass writes

The read answers for the state the previous pass left behind, which is what a plan is entitled to read. It costs nothing here: the controller's serving CA is persisted in the `unbounded-net-serving-cert` Secret with a ten-year lifetime, so a rollout does not change the value stamped into a registration. The CA ConfigMap watch and the 15 second poll close the remaining gap in one pass.

A failed backend read is a planning error, so net writes nothing that pass. That is the same consequence every component already accepts for a failed `resourcesExist` or `planConfig` read.

## Controller Readiness

The controller no longer publishes its Service endpoint immediately after acquiring leadership.

It waits until:

- It is the leader.
- Its informer caches have synchronized.
- The site controller has initialized.
- The CIDR allocator has been seeded.
- Endpoint publication succeeds.

The controller's `/readyz` endpoint reports Ready only after these conditions are satisfied.

Endpoint publication is retried on failure and periodically refreshed so deleted or drifted endpoint objects are repaired.

The same endpoint publication path is used when leader election is disabled.

## Live API Reads

Readiness decisions go through `mgr.GetAPIReader()` rather than the controller-runtime cache, exposed to components as `Env.LiveReader()`.

The cache is the wrong source twice over. It is populated by a watch, so it lags the API server and a rollout that has just started still reads as the settled previous one. And every kind read through it establishes an informer, so answering whether one Service has a backend would cache every Endpoints, EndpointSlice, Pod and ReplicaSet in the namespace for the life of the process.

`SetupWithManager` refuses a nil `APIReader` so a running operator never falls back to the cache.

## CA Handling

The operator reads the serving CA from the `unbounded-net-serving-ca` ConfigMap and stamps it into:

- Every validating webhook client configuration
- Every mutating webhook client configuration
- The aggregated `APIService`

Existing registrations are considered usable only when every bundle matches the currently published CA.

The operator watches the CA ConfigMap so new or rotated CA data is reconciled immediately rather than waiting for the polling interval.

The controller retains its existing startup CA injection for compatibility with older operators.

## Upgrade Behavior

During an expected `maxSurge: 0` controller rollout, existing usable registrations are left unchanged while the new revision starts.

The operator continues polling every 15 seconds so withheld changes are applied after recovery.

`NetReady` remains true when the existing registrations still have the correct CA. It becomes false with reason `BackendNotReady` when a required registration is missing or unusable.

## Not Gated

`ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` objects are still applied immediately.

These policies run inside the API server and do not depend on the net controller backend.


phealy (dismissed):
lgtm

plombardi89:
Review follow-ups are in. Two commits pushed:

- `101bf29b` net: keep the registration gate from wedging net itself
- `a6c75d64` net: fail the pod when the site controller cannot be created

**What changed**

1. **A failed backend read no longer aborts the plan.** `readBackendState` now returns a `backendState` with `readErr` set instead of an error, so a 403 on `endpoints`/`endpointslices` (for example during an upgrade whose ClusterRole has not landed yet) withholds the three registrations and lets the ConfigMap, Deployment, Service and node DaemonSet keep converging. The error is still carried into the Site condition and the aggregated reconcile error via the new `component.NotReadyErr`.
2. **The endpoint port is matched by name, not number.** `hasHTTPSPort` hardcoded 9999 while the controller publishes `cfg.HealthPort`, so any `controller.healthPort` override made the backend permanently unrecognisable and pinned `NetReady` False.
3. **A nil `targetRef` is accepted.** Controllers released before this PR do not set one, and a workload override can pin the controller image to such a version indefinitely, which froze the registrations silently (nothing pending, so `NetReady` stayed true while manifest changes were never applied).
4. **Two-tier poll.** 15s while a registration is actually missing or stale; 2m when everything is already in place with the current CA. Net is a cluster component planned once per Site request, so the old flat 15s re-applied the whole manifest set per Site forever behind a scaled-to-zero Deployment or a pinned image.
5. **`NewSiteController` failure is fatal.** It was logged and stepped over, which now produces a leader that holds the lease, passes its probes, publishes no endpoint, and keeps the operator withholding registrations for the life of the pod.
6. **Corrected the `readinessStatus` comment.** It claimed net is deployed before the machina CRDs exist; `BootstrapCRDs` applies and waits for every required CRD to be Established before the manager starts, so that cannot happen under the operator. It is true of the standalone `deploy-direct` path and of CRD-deleted recovery. The conclusion is unchanged.
7. **`ReconciledAfter` takes a message**, so the withholding-changes-nothing condition says why instead of "component reconciled".

New tests cover the read-failure path (workloads still applied, registrations withheld, `Err` propagated), a nil `targetRef` on both endpoint shapes, and a non-default port. `make fmt`, `make lint` and `go test -race` are clean on the touched packages.

Remaining review items are tracked in #628: the site-controller readiness deadline and `controller_ready` metric, making `controller.healthPort` genuinely configurable (or removing it), three small cleanups, and reference docs for `BackendNotReady`.

plombardi89:
@phealy I've squashed the review follow-ups so there's less to re-read. Two things changed since your lgtm on `3b016e8a`:

**1. [`59fa2096`](https://github.com/Azure/unbounded/commit/59fa2096c94de42e266bdd47e483a392c9846c50) — the merge with main.** Worth flagging rather than skipping: #587 landed after your review, so the net component had to be adapted to the plan-and-execute contract. That merge carries hand-resolved conflicts across `component/env.go`, `component/result.go`, `components/net/{activation,net,net_test}.go`, `reconciler{,_test}.go` and `cmd/unbounded-operator/main.go`. `git show --cc 59fa2096` is the readable view of just the resolutions.

**2. [`777fd4eb`](https://github.com/Azure/unbounded/commit/777fd4eb) — the response to review.** 518 lines across 8 files, one commit. It covers:

- A failed backend read no longer aborts the plan. Reading endpoints/endpointslices/pods/replicasets needs RBAC granted separately from what the operator applies with, so a 403 mid-upgrade used to withhold the ConfigMap, Deployment and node DaemonSet too. Now only the three registrations are withheld; the error still reaches the Site condition and the reconcile error.
- The endpoint port is matched by name, not number. `controller.healthPort` is user-settable, so a number there made any override pin `NetReady` False permanently.
- An endpoint with no `targetRef` is accepted. No released controller sets one, and a workload override can pin the image indefinitely, which silently froze the registrations with nothing reported.
- Two-tier poll: 15s while something is pending, 2m when everything is already in place with the current CA. Net is planned once per Site request, so the flat 15s re-applied the whole manifest set per Site indefinitely.
- `NewSiteController` failure is now fatal. Since publication waits on site-controller readiness, logging past it produced a leader that held the lease, passed its probes, and never published an endpoint.
- A comment pass: four comments described code that wasn't there (details in the commit message).

Remaining items from the review are tracked in #628 rather than growing this PR: the site-controller readiness deadline and a `controller_ready` metric, making `controller.healthPort` genuinely configurable or removing it, some cleanups, and reference docs for `BackendNotReady`.

The force-push only reshaped history — the tree is byte-identical to what CI last ran on `f1090ce8` (same tree hash `61857c16`), so the compare link on the force-push event is empty.

Separately: Vulnerability Check is red on this head. Looking into it now; I don't believe it's related to these changes.

plombardi89:
Follow-up on the red Vulnerability Check: **not related to this PR, and it will hit every open PR and main.**

Seven new advisories against `github.com/lib/pq` were published **today** (2026-08-18): GO-2026-6166, 6168, 6169, 6170, 6171, 6172, 6173. Example, [GO-2026-6173](https://pkg.go.dev/vuln/GO-2026-6173) / CVE-2026-56874:

> The special-case parser for PostgreSQL pre-protocol plain-text errors in github.com/lib/pq calls bufio.Reader.ReadString(0) without a byte limit. A malicious PostgreSQL endpoint or active network attacker on an unauthenticated transport can send an error message response with non-NUL bytes continuously, causing the client to buffer an arbitrarily large string and crash with an unrecoverable out-of-memory error.

All seven are **all versions, no known fix**. `v1.12.3` is already the latest release. They reach us through `cmd/inventory/inventory-aggregator`.

Why it fails now rather than before: the `vulncheck` target in the Makefile allowlists exactly one unfixable finding, and only when it is the only one:

```make
if grep -q 'affected by 1 vulnerability' "$$tmpf" && grep -q 'GO-2024-3218' "$$tmpf"; then
```

We are now at 8 (the 7 new lib/pq plus the existing libp2p GO-2024-3218), so the guard no longer matches. Main's last green Vulnerability Check was on `acb37bcc` at 2026-08-14, four days before these advisories existed.

Evidence it is unrelated to this branch: `git diff origin/main...HEAD -- go.mod go.sum` is empty, and nothing here touches the inventory or gantry packages.

This needs a maintainer decision rather than a drive-by fix in this PR, and I'd suggest a separate one:

1. Extend the allowlist to the seven lib/pq IDs with a written justification, in the style of the existing GO-2024-3218 note. The threat model is a hostile PostgreSQL endpoint or an on-path attacker on an unauthenticated transport, which may be out of scope for how the aggregator connects.
2. Or move off `lib/pq` (`jackc/pgx` is the usual replacement, and is maintained).
3. Or rework the guard to take a list of accepted IDs rather than matching on a count, so the next batch does not wedge CI the same way.

Happy to open an issue or send a PR for whichever you prefer.

phealy (approved):
lgtm

Issue #610, opened 2026-08-13
Release Machinery


plombardi89:
Capturing wants and requirements for release machinery. We're reaching a point on the maturity curve where we need to be confident that we can release at any point in time as well as support existing releases as necessary:

# Essentials

Mix of documentation, pipelines, and process:

1. How to verify we are stable for a release
2. How to pick a version (minor, patch, hotfix etc.)
3. How to prepare a release
4. How to promote a release
5. Hotfix releases from a release tag to support special use-cases.

plombardi89:
Status of the five essentials, after #619:

1. **How to verify we are stable for a release** — RELEASING.md §1, and the guides now open with it.
2. **How to pick a version** — RELEASING.md §2.
3. **How to prepare a release** — RELEASING.md "Cutting a release", with patch, minor and candidate-train walkthroughs.
4. **How to promote a release** — RELEASING.md "With a release candidate first". `promote` now tags the last candidate's commit rather than main HEAD, so the tree that was soaked is the tree that ships.
5. **Hotfix releases from a release tag** — still open, now tracked in detail as #627. The pipeline cuts from `main` only; the agreed model is a lazy `release-X.Y` branch with line-scoped version resolution, CI extended to `release-*`, and a maintenance release skipping the soak and staying a draft until published through the recorded break-glass path.

RELEASING.md documents the roll-forward remedy that exists today under "Shipping an urgent fix", including the one partial answer the pipeline does have: cutting a candidate the moment a fix lands holds that exact tree, since promote excludes anything merged afterwards.

plombardi89:
All five essentials are now in place.

| # | Essential | Where it lives |
|---|---|---|
| 1 | How to verify we are stable for a release | `RELEASING.md` §1, "Is `main` releasable?" |
| 2 | How to pick a version | `RELEASING.md` §2 plus the rule in #627: `main` cuts minors, `release-X.Y` cuts patches |
| 3 | How to prepare a release | `release-prepare.yaml`, one dispatch |
| 4 | How to promote a release | `mode=promote`, and the soak publishes automatically |
| 5 | Hotfix releases from a release tag | #627: `create-release-branch.yaml` plus series-scoped resolution |

`v0.4.0` was cut on 2026-08-24 and published through the full gate — resolver, build, soak, publish — with no manual step between the dispatch and the published release.

Two follow-ups are tracked separately rather than here:

[...]

PR #619, opened 2026-08-14
ci: make the release gate and version resolution trustworthy, and document the process

plombardi89:
## Summary

The release pipeline had accumulated several independent ways to either block a good release indefinitely or publish a release that had not actually passed the checks attributed to it. This PR hardens the rollout gate, release preparation, artifact verification, and publication path, and documents the resulting process in `RELEASING.md`.

### Unreachable nodes permanently blocked releases

A DaemonSet includes every eligible node in `desiredNumberScheduled`, even when the kubelet has stopped reporting. One unreachable edge node therefore left `gantry` and `unbounded-net-node` at 15/16 on `unbounded-stable`, causing every rollout gate to wait for its full timeout. Releases were consequently being published by hand, outside the automated gate and without an audit trail.

`hack/release/wait-rollouts.sh` now tolerates a DaemonSet shortfall only when it is fully explained by a bounded number of NotReady nodes. This path is strictly fail-closed because it can turn a failed rollout into a successful one. It requires all of the following:

- the workload is a DaemonSet;
- its current template contains the expected release tag on every project-published image;
- the number of NotReady nodes is within `MAX_NOTREADY_NODES`;
- the controller has observed the current generation and reports no misscheduled pods;
- every reachable owned pod is healthy and current;
- distinct healthy nodes plus distinct stranded nodes cover the desired fleet.

The gate counts nodes rather than pods, excludes terminating reachable pods, scopes evidence to pods controlled by the workload, and re-confirms the expected release before accepting either normal rollout success or a tolerated shortfall. Deployments remain outside tolerance because they can reschedule away from a failed node.

Both nightly and release-upgrade default to tolerating at most two NotReady nodes. Manual dispatch can tighten or override the limit, while the automatic release path always uses the workflow default. Diagnostics and smoke output now include node readiness and identify stranded pods by site.

### The gate could validate the wrong release

`kubectl rollout status` answers whether the current workload is settled. Before this PR, an old, healthy workload could report success before the operator had rewritten it for the release under test. Concurrent deployments could also replace workloads while another run was waiting, allowing one release to claim another release's rollout.

The gate now waits in three phases: the workload must exist, its template must reference the expected release, and that specific release must roll out. The expected tag is checked again after success. Deployments to each shared cluster are serialized with a cluster-scoped concurrency group so different tags cannot race on the same cluster.

The image check distinguishes project images by `EXPECTED_IMAGE_REGISTRY`. All project-published images must carry the expected tag, while intentionally pinned third-party images, such as gantry's busybox init container, are ignored. A stale primary container can no longer be hidden by a current sidecar.

### The soak did not cover everything it claimed to cover

Metalman runs on `unbounded-stable` but was absent from the rollout gate. It is a per-Site component, and its deployment names cannot be derived from the cluster's own `SITE_NAME`. `hack/release/metalman-targets.sh` now discovers every Site that enables Metalman and adds its Deployment to the release gate. Query failures and malformed cluster data fail closed rather than looking like an intentionally empty result.

Smoke discovery also had two false-green paths: an empty smoke directory skipped the matrix, and a failed discovery job also produced a skipped matrix. Publication accepted both. Discovery now fails when no smoke tasks are present, and normal publication requires both discovery and every smoke task to succeed. Nightly uses the same rule.

### Break-glass publication was unaudited and unsafe

When the soak cluster prevented a release from passing, maintainers published drafts manually with `gh release edit --draft=false`. That bypass left no durable explanation and made the release gate advisory in practice.

The release workflow now has a dedicated `publish-forced` path that:

- is reachable only through manual dispatch with `force_publish` and a written reason;
- uses the `unbounded-reviewers` environment so required reviewers can be configured independently;
- still runs artifact and provenance verification;
- records the actor, reason, and every gate result in the release body;
- composes the audit notice before publication and applies the notice and `draft=false` in one API call;
- does not publish after workflow cancellation.

Normal and forced publication are mutually exclusive for UI boolean inputs, REST/`gh` string inputs, and automatic `workflow_run` events.

### Release artifacts were incompletely verified

Artifact verification previously lived inline in one job and missed several failure modes. The shared `hack/release/verify-release-artifacts.sh` now protects both normal deployment and forced publication by checking:

- the manifest archive, operator manifest, and release BOM Sigstore bundles against this repository's tag-scoped release workflow identity;
- the BOM's release tag and commit;
- every image signature named by the BOM;
- a non-empty, completely parsed image list;
- every artifact and signature bundle declared by the BOM is still present in the GitHub Release;
- any binary about to be executed matches the signed `checksums.txt`.

This closes the path where the downloaded `kubectl-unbounded` plugin was executed against the cluster without verification, as well as partial-BOM parsing and incomplete-draft publication failures. The verification tooling is loaded from the workflow's own commit so backfills are not dependent on an older release tag containing current gate scripts.

### Version resolution could fork, guess, or tag the wrong tree

The old inline resolver reapplied `bump` to the latest final on every run, accepted unvalidated prerelease suffixes, guessed which train to promote, and tagged `main` HEAD rather than the candidate commit that had actually soaked.

Version resolution now lives in `hack/release/next-version.sh` and follows one consistent policy:

- an existing live train is continued instead of forked;
- blank `pre` selects the next canonical `rc.N` automatically;
- only bounded, non-zero-prefixed `vX.Y.Z-rc.N` tags count as release candidates;
- malformed, legacy, and off-branch tags do not create release trains or affect numbering;
- stale trains older than the latest final are reported but cannot be promoted;
- concurrent trains require an explicit opt-in, after which promotion requires an explicit version;
- version components and RC numbers are bounded to avoid Bash arithmetic overflow;
- promotion tags the highest canonical RC's commit, not current HEAD;
- the promoted commit must be an ancestor of the branch being released.

`dry_run` reports the tag and commit without pushing. Release preparation is serialized so two dispatches cannot race to mint the same version.

### Workflow errors and release knowledge were hard to detect

`release-prepare.yaml` contained a duplicate YAML key that made it undispatchable, but CI did not lint workflow syntax. `make lint` now runs pinned `actionlint`, with repository-specific runner labels configured.

`RELEASING.md` provides task-oriented guides for direct releases, candidate trains, promotion, recovery, break-glass operation, urgent fixes, and the current lack of maintenance branches. The large process comments previously duplicated across workflows were reduced in favor of this single runbook.

The nightly operator build also drops component build arguments that its Containerfile never consumed; component tags are selected from the operator's compiled version during reconciliation.

## Testing

The release shell tooling is exercised through committed tests that run the real scripts against controlled fake `kubectl`, Git repositories, `cosign`, `jq`, and checksum inputs. Coverage includes:

- healthy, degraded, stale, replaced, misscheduled, duplicated, foreign-owned, and malformed rollout states;
- Metalman discovery success, absence, query failure, and malformed payloads;
- forced-publication result reporting, required reasons, atomic notice composition, and idempotency;
- valid, missing, malformed, tampered, unsigned, and incomplete release artifacts;
- release, prerelease, concurrent-train, stale-train, off-branch, malformed-tag, numeric-boundary, and promotion-commit version cases.

The resolver suite currently passes 65 cases, and `go test ./hack/release` passes.

## Operational Note

Repository administrators should configure required reviewers on the `unbounded-reviewers` environment. GitHub creates the environment on first reference, so this does not block merging, but the forced path has no second-person approval until those protection rules are added.

Part of #610. Storage supervisor rollout coverage remains tracked in #625.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,12 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# actionlint configuration. Run via `make lint-actions`, which `make lint`
+# includes, so a malformed workflow is caught before it is dispatched rather
+# than when someone tries to cut a release with it.
+
+# Labels for self-hosted runners actionlint cannot know about. `pve` is the
+# Proxmox host pool used by aks-ssh-quickstart.yaml.
+self-hosted-runner:
+  labels:
+    - pve
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -405,7 +405,18 @@ jobs:
       SITE_NODE_CIDR: ${{ vars.SITE_NODE_CIDR }}
       SITE_POD_CIDR: ${{ vars.SITE_POD_CIDR }}
       MANAGE_CNI_PLUGIN: ${{ vars.MANAGE_CNI_PLUGIN }}
-      MAX_NOTREADY_NODES: ${{ inputs.max_notready_nodes || '0' }}
+      # Same cap as the release deploy gate. The nightly deploys the same
+      # component images to the same shape of cluster, and a node dropping out
+      # of unbounded-nightly says no more about this snapshot than one dropping
+      # out of unbounded-stable says about a release. A stricter cap here would
+      # only mean the nightly goes permanently red the first time a remote site
+      # goes offline, which is the failure this whole check exists to end.
+      #
+      # `||` relies on expression truthiness, which is safe here only because
+      # the input is typed `string`: a non-empty string, including "0", is
+      # truthy. Retyping it to `number` would silently turn an explicit 0 back
+      # into this default.
+      MAX_NOTREADY_NODES: ${{ inputs.max_notready_nodes || '2' }}
       # The nightly image tag, which is also the version stamped into the
       # operator binary, so it is the tag the operator resolves every component
       # image at. wait-rollouts.sh refuses to excuse a shortfall on a workload
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -388,6 +388,11 @@ jobs:
       # every component image at. wait-rollouts.sh will not excuse a shortfall
       # on a workload that does not carry it yet.
       EXPECTED_IMAGE_TAG: ${{ needs.resolve.outputs.tag }}
+      # Where we publish. The gate needs it to tell the images this release
+      # ships from the third-party ones pinned beside them, such as gantry's
+      # busybox init, which never carries a release tag. Lowercased by the
+      # script, since the owner is not.
+      EXPECTED_IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
     outputs:
       mode: ${{ steps.mode.outputs.mode }}
     steps:
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -315,14 +315,12 @@ jobs:
 
       - name: Compute per-component build args
         # metalman bakes its default netboot image from CONTAINER_REGISTRY, so
-        # it needs that arg. REGISTRY is lowercased by the step above.
+        # it needs that arg (images/metalman/Containerfile declares it and feeds
+        # it to make). REGISTRY is lowercased by the step above.
         #
-        # The four unbounded-operator args are INERT: images/unbounded-operator/
-        # Containerfile declares only TARGETOS/TARGETARCH/VERSION/GIT_COMMIT/
-        # BUILD_TIME, so docker ignores them. Nothing breaks, because the
-        # operator overwrites every manifest image at reconcile time from its
-        # own compiled version (internal/operator/component/env.go) - which is
-        # also why VERSION is the arg that actually matters.
+        # Nothing else needs one. Which component images the operator deploys is
+        # decided at reconcile time from its own compiled version, so VERSION,
+        # passed to every build below, is what selects them.
         run: |
           set -euo pipefail
           {
@@ -330,12 +328,6 @@ jobs:
             case "${{ matrix.component.name }}" in
               metalman)
                 echo "CONTAINER_REGISTRY=${REGISTRY}"
-                ;;
-              unbounded-operator)
-                echo "NET_CONTROLLER_IMAGE=${REGISTRY}/unbounded-net-controller:${TAG}"
-                echo "NET_NODE_IMAGE=${REGISTRY}/unbounded-net-node:${TAG}"
-                echo "MACHINA_IMAGE=${REGISTRY}/machina:${TAG}"
-                echo "STORAGE_SUPERVISOR_IMAGE=${REGISTRY}/unbounded-storage-supervisor:${TAG}"
                 ;;
             esac
             echo 'EOF'
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -406,6 +406,11 @@ jobs:
       SITE_POD_CIDR: ${{ vars.SITE_POD_CIDR }}
       MANAGE_CNI_PLUGIN: ${{ vars.MANAGE_CNI_PLUGIN }}
       MAX_NOTREADY_NODES: ${{ inputs.max_notready_nodes || '0' }}
+      # The nightly image tag, which is also the version stamped into the
+      # operator binary, so it is the tag the operator resolves every component
+      # image at. wait-rollouts.sh refuses to excuse a shortfall on a workload
+      # that does not carry it yet.
+      EXPECTED_IMAGE_TAG: ${{ needs.resolve.outputs.tag }}
     outputs:
       mode: ${{ steps.mode.outputs.mode }}
     steps:
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -6,28 +6,20 @@ name: nightly
 # ---------------------------------------------------------------------------
 # Builds a snapshot of `main` from source every morning and deploys it to the
 # integration cluster (unbounded-nightly) so the tip of the tree gets the same
-# soak treatment that releases get on unbounded-stable.
+# soak treatment releases get on unbounded-stable.
 #
-# Relationship to release-upgrade.yaml
-# ------------------------------------
-# This is the nightly sibling of .github/workflows/release-upgrade.yaml. That
-# workflow deploys a PUBLISHED, signed release (downloaded as a manifests
-# tarball from a GitHub Release) to unbounded-stable. This one has no release
-# and no tag: it builds the current main HEAD from source, pushes images
-# tagged `nightly-<shortsha>`, then bootstraps the operator via
-# `kubectl unbounded install` and lets it reconcile the Site into
-# unbounded-nightly (same operator-driven flow as release-upgrade.yaml). There
-# is deliberately no cosign signing, SBOM, or Trivy gating here: the nightly
-# cluster is a throwaway soak target, not a customer artifact. The full supply
-# chain stays on the release path.
+# The nightly sibling of release-upgrade.yaml, which deploys a published signed
+# release to unbounded-stable. This one has no release and no tag: it builds
+# main HEAD, pushes images tagged `nightly-<shortsha>`, and lets the operator
+# reconcile them in, using the same deploy gate and smoke tests. Deliberately no
+# cosign, SBOM or Trivy: the nightly cluster is a throwaway soak target, and the
+# supply chain stays on the release path.
 #
 # Schedule
 # --------
-# Runs at 06:00 UTC daily. GitHub Actions cron is UTC-only and does not follow
-# US daylight saving, so this lands at 01:00 ET (EST) / 02:00 ET (EDT). We run
-# early on purpose: a from-source build of every component plus a deploy and
-# smoke pass needs to finish well before the US working day, so 06:00 UTC is
-# chosen over a literal 4am-ET cron to leave that headroom.
+# 06:00 UTC daily, which is 01:00/02:00 ET depending on daylight saving (Actions
+# cron is UTC-only). Early on purpose: a from-source build of everything plus a
+# deploy and smoke pass has to finish before the US working day.
 #
 # Target cluster
 # --------------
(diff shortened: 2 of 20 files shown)


plombardi89:
Pushed eight commits addressing review. The PR body has been rewritten to match; the substantive additions are §2 and §6, which are new behaviour rather than fixes to what was already described.

**Blocking issues**

- `release-prepare.yaml` carried a duplicate `default:` key (a stray line from the `version` input rewrite, which also took the default off `version` itself). GitHub rejects duplicate mapping keys, so the workflow that mints every version tag could not be dispatched. Fixed, and `make lint` now runs actionlint so this class cannot reach green CI again - it found this in under a second. Only one pre-existing finding repo-wide, the `pve` runner label, now declared in `.github/actionlint.yaml`.

- Tolerance could excuse the **previous** release. Component workloads are updated asynchronously, so for the first seconds of a wait a DaemonSet still carries the old template - and every other condition holds in that window. On `unbounded-stable`, with its one permanently unreachable node, the gate would have returned success about a second after starting, having validated nothing, and `publish` would then have shipped it automatically with no bypass recorded. Tolerance now also requires the workload's current pod template to reference the release's image tag. `TestRefusesToTolerateThePreviousRelease` is the guard; with the condition removed it reports `OK: all workloads rolled out` for a workload on the old tag.

**Also**

- `promote` tagged `main` HEAD, so `v0.2.5` shipped a different tree than the `v0.2.5-rc.3` that was soaked. It now resolves the candidate's commit, with guards for a train with no `rc` and for a base that is not an ancestor of HEAD.
- Both scripts now have committed tests that CI runs: 29 cases for `wait-rollouts.sh` (12 new) and 39 for `next-version.sh`, the latter previously executed by nothing at all.
- Nightly cap raised to 2 to match the release gate, rather than correcting the runbook to match the cap - the runbook described the behaviour we want, and a stricter cap only meant the nightly goes permanently red the first time a remote site drops out.
- Nits: deterministic `describe_nodes` ordering, conditional and idempotent forced-publish notice, escaped tag in the cosign identity pattern, and `core-namespaces-ready.sh` no longer passes vacuously when its pod query fails.

**Deliberately not here:** #623 (shellcheck over `run:` blocks), #624 (the image guard has the same staleness the tolerance check just fixed), #625 (the gate waits on 5 of 7 component workloads).

Two things still cannot be exercised before merge, both now called out in the body: the forced-publish path runs for the first time on first use, and `release-prepare` reads the resolver from `main`, so no resolver change - `dry_run` included - can be rehearsed by dispatching this branch.

plombardi89:
Three more commits, all green.

**1. Metalman is now required** (`72337ae5`)

The targets could not be built from `SITE_NAME`. Metalman is a per-Site component, and on unbounded-stable the environment's `SITE_NAME` is `stable` while metalman runs for a remote site — gating on `metalman-controller-stable` would have waited 300s for a workload that does not exist and failed every release. `hack/release/metalman-targets.sh` discovers `deploy/metalman-controller-<site>` for every Site that enables it.

It is a script rather than YAML because its empty answer silently removes metalman from the gate. So it separates *no Site enables metalman* (exit 1) from *I could not tell* (exit 3) — an unreachable apiserver, an unexpected payload shape or a jq error never reads as "none enabled". Eight cases cover that, including both jq failure modes.

Consequence you accepted explicitly, and which is in the commit message: metalman is a Deployment, Deployments are outside node tolerance, so if the site it runs for has no Ready node the release gate goes red.

**2. Smoke can no longer be skipped** (`472adba7`)

You were right to push on this, and it was worse than the question implied — two paths, not one:

- `hack/release/smoke/` missing or empty **on the default branch** → `has_tasks=false` → skipped → published. Since discovery checks out main rather than the tag, that is a repo regression, not a property of the release: deleting or renaming that directory would have disabled smoke for every future release.
- `smoke-discover` **failing** → matrix skipped → `result == 'skipped'` → published, because `publish` never checked that job's result. A broken discovery step shipped a release having run nothing.

Discovery now fails instead of emitting `has_tasks=false`, which makes the matrix impossible to empty and lets both the output and the `if:` gate go. `publish` requires `smoke-discover` and `smoke-tests` to have succeeded. Skipping is only reachable through `publish-forced`, where it is recorded — the manual override you asked for already existed.

I also fixed the same branch in the nightly, where the result was a green nightly that had validated nothing. That is outside the "stable only" scoping, which was about metalman — say the word and I will split it out.

**3. Inert build-args removed** (`507555af`, closes #626)

The `Makefile` `ifeq/override` guards stay: they exist precisely because these once *were* declared `ARG`s, and Docker exporting them empty blanked the images baked into the embedded manifests.

#625 is now storage-only.

jwilder (approved):

jwilder (dismissed):

plombardi89:
All seven fixed, seven commits, 24/24 green. Every fix has a test that fails when the fix is reverted — I ran each mutation rather than asserting it.

| # | Commit | Guard |
|---|---|---|
| 1 | `b1d2f9e4` | `TestRefusesAHealthyWorkloadOnThePreviousRelease` |
| 2 | `54218da4` | `TestRefusesToTolerateAnOutdatedPodOnAReachableNode` |
| 3, 6 | `895e64fb` | `TestForcedNotesDoNotClaimASoakThatWasSkipped`, `TestForcedNotesAreIdempotent` |
| 4 | `bae64357` | `TestVerifierRejectsATamperedBinary` |
| 5 | `74006143` | `TestVerifierRejectsAPartialImageList` |
| 7 | `cf68c7cd` | 7 cases; `rc.08` and a stray `v1.2` both mint tags without the fix |
| — | `86605fb9` | 15 verifier cases, replacing a matrix that only ever existed in this description |

Three were worse than reported, which changed how I fixed them:

**#2 is not a race.** With `maxUnavailable: 1` a stranded pod counts as unavailable, so the DaemonSet controller *stops* updating the remaining nodes. "Unreachable node's pod updated, reachable ones not" is the steady state of a stalled rollout. At desired 16 it tolerated with a whole node still on the old release. Patching the arithmetic would have left the same class open, so every count now comes from the pod list and no controller counter is added to a pod-derived number.

**#7 fails silently, not loudly.** `(( want <= have ))` with `rc.08` sits in an `if` condition, which `set -e` exempts, so the octal error makes the comparison false and the "must be ahead" guard never fires — `rc.08` was accepted *behind* `rc.5`, and the tag it minted then poisoned `max_rc` the same way. Similarly a stray `v1.2` produced `v1.2.1`, which passes every later check because it looks well formed.

**#4 needed no new signing.** GoReleaser already signs `checksums.txt`, so `checksums.txt.bundle.json` ships under the same `release.yaml@refs/tags` identity. The chain existed and was unused. `CHECKSUM_FILES` names what a caller is about to execute; entries must be present, listed in the signed file, and matching — the listing check matters because `--ignore-missing` would otherwise pass over an artifact absent from the list.

Two notes on judgement calls:

- Fixing #1 means a component pinned by an `unbounded-component-overrides` entry will now time out the deploy rather than pass. You chose hard-fail; the error names the override as the likely cause.
- Extracting the forced-publish notice found a third defect the review did not: the separator differed depending on whether the generated changelog ended in a newline. Fixed and covered.

The PR description's "Stubbed-cosign matrix … 7/7" claim is gone — that was the gap you identified, and it is now 15 committed cases.

plombardi89:
All six fixed, six commits, 24/24 green. Each fix has a test that fails when the fix is reverted — I ran every mutation rather than asserting it. The two exceptions are noted below.

| # | Commit | Guard |
|---|---|---|
| 1 | `5a865334` | none possible — workflow-level condition |
| 2 | `058647a2` | `TestRefusesAWorkloadReplacedDuringTheRollout`, `…CannotBeReReadAfterRollout` |
| 3 | `56010596` | `TestRefusesToTolerateWhenANodeIsCoveredTwice`, `TestIgnoresPodsThisWorkloadDoesNotOwn` |
| 4 | `92d70d43` | `TestRefusesAStalePrimaryBesideACurrentSidecar`, `TestAcceptsAPinnedThirdPartyImage` |
| 5, 6 | `a6a83f9e` | 6 cases; an off-branch `v9.0.0` mints `v9.0.1` and a 20-digit tag is bumped without them |

**#1** — `publish` drops `always()` rather than swapping it for `!cancelled()`. Its four `needs.*.result == 'success'` checks already say exactly what default job semantics enforce, so the function bought nothing and cost cancellation. `publish-forced` genuinely needs one, since it runs when deploy failed, so it takes `!cancelled()`. I also swept the two summary steps and the commented-out notify template in that file — leaving three uses of the pattern next to jobs that just moved off it invites the next person to copy the wrong one. `always()` elsewhere in the repo is untouched: those workflows publish nothing.

**#2** — I took both halves. Serializing per cluster is the root fix; the tag re-check after rollout is defence in depth, because a shared cluster has writers that are not this pipeline. Worth knowing: a manual dispatch, including a break-glass re-run, now queues behind an in-flight automatic deploy. That is the cost of the gate having observed one specific thing.

**#3** — the owner scoping refuses when the workload has no uid rather than skipping the filter. It cannot happen against a real apiserver, and a fail-open there would defeat the point.

**#4** — `EXPECTED_IMAGE_TAG` and `EXPECTED_IMAGE_REGISTRY` are now required together, with a hard exit if only the tag is set. A quiet fallback to the old `any` rule is how the sidecar hole appeared, and it would reappear the moment a third caller set one and not the other. Consequence: a cluster whose operator points at a different registry fails the gate until the variable matches.

**#5** — I checked the effect on this repository before changing it, since `--merged HEAD` was the option I rejected last round. Ten tags are unreachable from main, but none is the latest final, so `latest_final` is `v0.2.4` either way and only the stale-train report changes. Verified against the live tag set: `prerelease` still resolves `v0.2.5-rc.1`.

Test totals: 73 Go cases in `hack/release/`, 52 resolver cases.

plombardi89:
All five fixed, five commits. Each has a test that fails when the fix is reverted; I ran every mutation.

| # | Commit | Guard |
|---|---|---|
| 1 | `0e204a1a` | `TestRefusesAToleratedWorkloadReplacedMidEvaluation` |
| 2 | `3c68ed18` | `TestVerifierRejectsAReleaseMissingADeclaredArtifact`, `…ReportsEveryMissingArtifactAtOnce` |
| 3 | `ee55f9c2` | `TestRefusesToTolerateWithAMisscheduledPod` |
| 4, 5 | `db6b1c81` | 5 cases; without them a bump mints `v1000000000.0.0` and a doomed train starts |

**#2** — I quantified the gap before fixing it. The BOM declares six artifacts; verification covered three, plus `checksums.txt` on the deploy path only. I also pulled `checksums.txt` from v0.2.4: 16 entries, all GoReleaser binaries, so there was no indirect coverage of the storage tarballs or `unbounded.yaml` either. Per your steer the check is presence-based — every declared artifact and signature bundle must appear in the release's asset list, compared by name against `gh release view --json assets`, no downloads. It is required rather than optional: a BOM that declares artifacts with no asset list supplied is an error, since skipping quietly is how the gap survived. A BOM with no artifact list at all is an older release being backfilled, which reports and continues.

**#4** — this was three patterns disagreeing, not one bug: discovery bounded the digits, the promote input did not, and the final check on the computed tag did not either. They are now one `SEMVER_COMPONENT` composed into every form, and `bump_version` refuses an overflow itself rather than leaving the final check to report "not vX.Y.Z" about something that plainly is.

**#5** — worse than "cannot be promoted", which is worth recording: because the core stays permanently outside `LIVE`, every subsequent run reported *"Starting a new train"* and minted another rc. Verified in a fixture before fixing.

**#1** — for the record, the window here was the node and pod queries, not the minutes the rollout path exposes. It is closed anyway; "the window is small" is how the first version of this check ended up covering only the degraded path.

---

**One check is red and it is not this PR.** `Vulnerability Check` fails on four `github.com/lib/pq` advisories published today at 16:38Z (GO-2026-6170/6171/6172/6173), reachable via `internal/inventory`, with no fixed version available. I ran govulncheck against a clean worktree of `origin/main` and it fails there identically; main's last CI run predates the advisories. Filed as #629 with the reachability traces and the options. The other 23 checks pass.

cchildress (approved):

PR #646, opened 2026-08-21
Add token-refresher

cchildress:
Adds token-refresher as an unbounded component. It will watch secrets in kube-system and regenerate the bootstrap token if it expires. We update any bootstrapTokenRefs in machine resources to match as well.


jveski:
Can this just be a controller compiled into the unbounded operator? Seems like it's small/simple enough to avoid adding another binary.

Also we might consider switching to short-lived service account tokens to avoid the need for long lived tokens in the first place.

cchildress:
@jwilder 

> Should we look for tokens that are about to expire vs already expired and rotate at some threshold? Then gradually update refs to the new?

Yup, I think that's a good idea. I don't really know what batch size to set yet, but I've at least done the early renewal.

> The PR description says it updates Machines but I don't see that code.

Only because I forgot to commit it...

@jveski 

>Can this just be a controller compiled into the unbounded operator? Seems like it's small/simple enough to avoid adding another binary.

I'm not opposed to it. I did this because I didn't want to start overloading the operator with extra functions. I view this as a place where other tokens / short-lived objects could be refreshed besides just bootstrap tokens in particular. Plus I'm adding a toggle so you can just disable this if needed. Let me know that rationale makes sense to you or if this is still just too much extra weight.

> Also we might consider switching to short-lived service account tokens to avoid the need for long lived tokens in the first place.

Sure? I'm going to say that's out of scope for now. Both token-refresher and machina mint tokens right now. I think what you're describing is a config directive to set the token duration. That can be another PR in the future.

cmd/token-refresher/tokenrefresher/controller/token_controller.go
```diff
@@ -0,0 +1,105 @@
...
+	token.WithLabel(unboundedv1alpha3.MachineSiteLabelKey, site.Name)
+
+	if err := kube.ApplyBootstrapToken(ctx, r.KubeClient, fieldManager, token); err != nil {
+		return ctrl.Result{}, fmt.Errorf("apply bootstrap token for Site %q: %w", site.Name, err)
+	}
+
+	ctrl.LoggerFrom(ctx).Info("Created bootstrap token", "site", site.Name, "expiresAt", token.ExpiresAt)
+
+	return ctrl.Result{RequeueAfter: time.Until(token.ExpiresAt)}, nil
+}
```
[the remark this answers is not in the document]
cchildress: Solved now. I'd forgotten to commit a section.

cmd/token-refresher/tokenrefresher/controller/token_controller.go
```diff
@@ -0,0 +1,267 @@
...
+
+func (r *TokenReconciler) reconcileMachines(ctx context.Context, siteName, desiredSecretName, rotatingSecretName string) error {
+	machines := &unboundedv1alpha3.MachineList{}
+	if err := r.List(ctx, machines, client.MatchingFields{machineSiteField: siteName}); err != nil {
+		return fmt.Errorf("list Machines for Site %q: %w", siteName, err)
+	}
+
+	var errs []error
+
+	for i := range machines.Items {
```
jwilder: I'm still a bit concerned about this when machines.Items is 100k.  The initial list above will be expensive and slow and the subsequent `ensureMachineTokenRef` below will be expensive too.  When this runs, we could see a big spike spike in memory/cpu and load on API server/etcd.  

Can we switch this so each reconcile does a smaller fixed batch of work, say 1000 machines?  I'm not sure how frequently, reconcile runs but smaller incremental work over a longer period should be less taxing then one huge reconcile all at once.

jwilder (changes requested):

jveski (dismissed):
LGTM - we'll just need to address the hot update loop at some point in the future

cmd/token-refresher/tokenrefresher/controller/token_controller.go
cchildress: Yup, done.

jwilder (approved):

PR #652, opened 2026-08-21
fix(gantry): route stalled pulls to next-ranked node

matucker-msft:
## Summary

Fix designated-puller takeover when the highest-ranked Gantry node has a stalled in-flight pull.

## Design deviation

The previous implementation did not match the takeover behavior specified in `designs/gantry-detailed-design.md`.

The design requires a puller whose in-flight operation exceeds the per-digest stall threshold to be excluded from subsequent `please_pull` candidates. The next-ranked reachable node should then take over.

Previously, Gantry recognized stale rank 0 during the in-flight piggyback check and incremented `p2p_designated_puller_takeover_total`, but Rule 7 selected from the original, unfiltered reachable set. This could send `please_pull` back to stale rank 0 instead of rank 1. The metric could therefore report a takeover that never occurred.

## Changes

- Track pullers whose in-flight work has exceeded the per-digest stall threshold.
- Exclude those pullers when Rule 7 selects the lowest-ranked eligible node.
- Preserve stalled-node responses for earlier failure, cache, and cooldown evaluation.
- Strengthen `TestRule3_InFlightStaleExcluded` to verify that `please_pull` targets the next-ranked node.

This brings the implementation back into conformance with the documented stalled-puller takeover flow.

## Validation

- `go test ./internal/gantry/coldstart ./internal/gantry/mirror ./cmd/gantry -count=1`
- `golangci-lint run ./internal/gantry/coldstart`

The new recipient assertion failed against the previous implementation because it selected stale rank 0. It passes after the fix and confirms that rank 1 receives the takeover request.

### DIFF
--- a/internal/gantry/coldstart/coldstart.go
+++ b/internal/gantry/coldstart/coldstart.go
@@ -549,7 +549,8 @@ func (r *Resolver) probe(ctx context.Context, d digest.Digest, kind ifaces.Origi
 	}
 
 	// Rule 3: in-flight piggyback.
-	if v := findInFlight(reachable, r.opts.Inflight, kind, expectedSize, r.opts.Now(), r.opts.Metrics.OnDesignatedPullerTakeover); v != nil {
+	inFlight, stalledPullers := findInFlight(reachable, r.opts.Inflight, kind, expectedSize, r.opts.Now(), r.opts.Metrics.OnDesignatedPullerTakeover)
+	if inFlight != nil {
 		if r.opts.Metrics.OnTopKProbeHit != nil {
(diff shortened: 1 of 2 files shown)


vpatelsj (approved):
this is a good find. Although it may seem otherwise, this bug only affects pulling a digest in the cold pull. We have prefetch cordinator that will perform a "please-pull" operation on top ranked HRW nodes automatically. So even if a Rank0 node was stalled , the other ranked nodes will start pulling that manifest/layers and populate the DHT. The node needing the actual payload will detect that Rank1 node has downloaded the payload and will start pulling it from Rank1 node while the Rank0 node stays stalled for whatever reason.

PR #653, opened 2026-08-21
release: tolerate workloads stranded by an unreachable site

plombardi89:
`unbounded-stable` has seven `boulderlab` nodes whose kubelets have stopped reporting. The release gate has failed every run since, so `v0.4.0` could only ship via `publish-forced` — which validates nothing, because `deploy` *fails* rather than skips and the smoke jobs never run.

Two checks judged those nodes wrongly.

## The problem

**`wait-rollouts.sh` tolerated a DaemonSet shortfall but never a Deployment's**, on the grounds stated in its header: *"a Deployment reschedules off a dead node"*. That stops being true when required affinity pins it to a site whose nodes are all gone — which is exactly what the operator creates, one per Site that enables a component:

```
0/16 nodes are available: 7 node(s) didn't match Pod's node affinity/selector,
9 node(s) had untolerated taint(s).
```

So the gate burned its full five-minute timeout on `metalman-controller-boulderlab`.

**`core-namespaces-ready.sh` then failed the same pod.** It classifies a stranded pod by looking up its node's readiness, and an unscheduled pod has no node — so it could never be classified as stranded, however dead its site was. That contradicts the convention at the top of that file, which says not to fail on workloads a site outage explains because the deploy gate has already decided.

## The solution

Both now excuse a workload on one rule: **every site it is pinned to has at least one node, and none of those nodes is Ready.**

Deliberately still refused:

| case | why |
|---|---|
| no site pin | an ordinary Deployment reschedules; a shortfall is real |
| a pinned site with a reachable node | it could be running there |
| a pinned site no node claims | a label matching nothing is a typo, not a dead site |
| preferred-only affinity | the scheduler may ignore it, so it could have run elsewhere |
| a term naming no site | terms are OR-ed, so it could have run wherever that term matches |

The gate additionally keeps its existing conditions: the workload must be on the release under test, within `MAX_NOTREADY_NODES`, and have no pod on a reachable node.

## Verification

Against the live cluster, the whole chain now passes, and the gate tolerates in **3.4s** where it previously ran to timeout:

```
1. deploy gate            exit=0
2. deploy-orca            exit=0
3. smoke core-namespaces  exit=0
```

Fourteen gate tests and thirteen smoke tests, all mutation-tested — removing any of the three new invariants turns tests red.

## Notes for reviewers

**gantry is fixed outside this PR.** It was the failure that actually tripped CI, before metalman was reached: `maxUnavailable` is a fleet-wide budget and pods on unreachable nodes count against it, so at `10%` of sixteen nodes seven dead nodes consumed it before the rollout began. An earlier revision raised that default and is reverted in `bf2d6d82` — `10%` is right for a healthy fleet, and users shouldn't inherit a change that exists to serve our soak cluster. `unbounded-stable` uses a component override instead, already applied.

**This must merge before a release run can pass.** The workflow takes `wait-rollouts.sh` and `hack/release/smoke/` from the default branch, not the deployed tag.

**The site logic is duplicated on purpose.** Smoke tests are meant to stand alone, so the two copies are kept structurally identical and each points at the other. That doesn't prevent drift, only makes it visible by diff; a shared helper is the alternative if you'd rather.

**Review found a fail-open in the first revision** — the OR case in the table above — now closed in both copies with a regression test. See `e930c1df`.

### DIFF
--- a/deploy/gantry/daemonset.yaml.tmpl
+++ b/deploy/gantry/daemonset.yaml.tmpl
@@ -82,22 +82,12 @@ spec:
   updateStrategy:
     type: RollingUpdate
     rollingUpdate:
-      # 100%, matching unbounded-net-node and unbounded-storage-supervisor.
-      #
-      # A percentage here is a budget for the WHOLE fleet, not just the nodes
-      # being updated, and pods on unreachable nodes count against it. At 10%
-      # of a 16-node fleet the budget was 1, so seven nodes that had lost their
-      # kubelet consumed it before the rollout started and the controller never
-      # updated any of the nine reachable nodes. The release gate then sat
-      # through its full timeout every time, because a fleet where most healthy
-      # nodes still run the previous release is not one it can certify.
-      #
-      # Restarting every agent at once is cheap here for the reason the old
-      # comment gave for keeping the number small: the mirror endpoint is
-      # per-node and containerd falls back to origin on connection failure, so
-      # an agent that is briefly gone only affects pulls happening in that
-      # window, and it costs latency rather than correctness.
-      maxUnavailable: 100%
+      # Limit blast radius: at most 10% of nodes lose their agent at
+      # any one time during an update. The mirror endpoint is per-node
+      # and containerd falls back to origin on connection failure, so
+      # a node losing its agent only affects pulls happening during
+      # the window.
+      maxUnavailable: 10%
   template:
     metadata:
       labels:
--- a/deploy/gantry/daemonset.yaml.tmpl
+++ b/deploy/gantry/daemonset.yaml.tmpl
@@ -82,12 +82,22 @@ spec:
   updateStrategy:
     type: RollingUpdate
     rollingUpdate:
-      # Limit blast radius: at most 10% of nodes lose their agent at
-      # any one time during an update. The mirror endpoint is per-node
-      # and containerd falls back to origin on connection failure, so
-      # a node losing its agent only affects pulls happening during
-      # the window.
-      maxUnavailable: 10%
+      # 100%, matching unbounded-net-node and unbounded-storage-supervisor.
+      #
+      # A percentage here is a budget for the WHOLE fleet, not just the nodes
+      # being updated, and pods on unreachable nodes count against it. At 10%
+      # of a 16-node fleet the budget was 1, so seven nodes that had lost their
+      # kubelet consumed it before the rollout started and the controller never
+      # updated any of the nine reachable nodes. The release gate then sat
+      # through its full timeout every time, because a fleet where most healthy
+      # nodes still run the previous release is not one it can certify.
+      #
+      # Restarting every agent at once is cheap here for the reason the old
+      # comment gave for keeping the number small: the mirror endpoint is
+      # per-node and containerd falls back to origin on connection failure, so
+      # an agent that is briefly gone only affects pulls happening in that
+      # window, and it costs latency rather than correctness.
+      maxUnavailable: 100%
   template:
     metadata:
       labels:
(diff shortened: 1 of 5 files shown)


plombardi89:
Thanks — finding 1 is a real fail-open and it was mine. All four substantive findings reproduced before acting on any of them.

## 1. The OR under-approximation — fixed, both copies

Reproduced against the gate's own filter:

```
site:bl + arch:amd64  =>  ["bl"]
```

The part that bothers me most is that the comment asserted *"exact for the OR"* while being exactly that. A wrong comment above a subtle filter is worse than no comment, because it stops the next reader checking.

Every `nodeSelectorTerm` must now contribute a site value, else the filter yields `[]`. Verified it doesn't over-refuse:

```
site:bl + arch:amd64 (the OR hole)   => []          <- now refuses
single term site:bl                  => ["bl"]
two terms, both site-pinned          => ["bl","edge"]
nodeSelector only                    => ["bl"]
real metalman deployment             => ["boulderlab"]   <- unchanged
```

Also fail-closed for `matchFields`-only terms and `NotIn` on the site key, which you didn't raise but land in the same place.

Your severity split was right, and it's why this mattered: the gate's `on_ready == 0` usually covers it, escaping only when the pod is *also* unschedulable for another reason — precisely when a genuine scheduling failure gets excused as a dead site. Smoke had no backstop at all.

## 2. Coverage — four gate tests, and I reversed the no-tests call

Confirmed: every `siteDeployment` call passed `nil` or a single-element slice, `net.unbounded-cloud.io/site` appeared zero times, and the two `nodeSelector` hits were `nodeSelectorTerms` in the builder rather than the branch.

Added the three you asked for plus a regression guard for the OR hole itself, so it can't silently reopen.

On the smoke tests: you changed my mind. "Your reasoning holds, but my thirty-line harness caught the OR hole immediately" is a hard argument to answer, and the evidence wasn't available when the call was made. Thirteen cases, reusing the fake kubectl the other shell tools here already use rather than a second harness.

## 3. Comment drift — fixed

The GET sits six lines above the comment claiming nothing had been queried. Corrected all three to claim what the tests actually assert: no node list.

I did consider making the claim *true* instead — `resolve_target` already reads the object pre-wait and could return the pins as a fourth line, restoring the genuine zero-call refusal. I didn't, because the pin can change mid-wait so the tolerate path needs the fresh read anyway, leaving two sources of truth for one fact to save a few KB every five seconds. Your call was the better trade; flagging only that "make it true" was weighed.

## 4. Luck, not shape — you're right

```
OLD 4-field: name=metalman-abc phase=Pending ready=spark-3d37 node=[]
NEW sep:     name=metalman-abc phase=Pending ready=[]         node=[spark-3d37]
```

So the old format lost the nodeName for a scheduled pod with no Ready condition, and reported a pod stranded on a dead node as a failure. Fail-closed and rare. My change fixed it by accident and I then wrote a comment asserting it couldn't happen.

Corrected in the comment and in the commit message — amended and force-pushed with the author's explicit sign-off, since this repo's conventions require that.

Best part: reverting the separator now fails `TestSmokeToleratesAnUnreadyPodOnANotReadyNode`, so that latent bug is covered rather than merely fixed.

## 5. Took all three

- **Named the dropped conditions.** `MAX_NOTREADY_NODES` and the image tag, plus your point that "already decided by the gate" spans a time gap and is an assumption rather than an invariant. Written down where it's checkable.
- **Structural parallelism.** The gate now uses `group_by` and grew the same `site_entirely_unreachable` helper, so the two are diffable rather than comparable-by-reading. You were right that this was the enforceable half.
- **Smoke's no-nodes warning.** Now says `pinned to site X, which has no nodes` like the gate.

The `jq` guard being inherited by every copy of the template is a fair observation. Leaving it: the template is the documented starting point and a copy that drops jq can drop the guard.

## Verification

Mutation-tested all three new invariants — every-term-names-a-site, all-nodes-Ready, and the unit separator — each turns tests red.

Re-ran the whole chain against `unbounded-stable` after the change, since a stricter filter could refuse something previously tolerated:

```
1. deploy gate            exit=0
2. deploy-orca            exit=0
3. smoke core-namespaces  exit=0
```

`make fmt`, `make lint`, full `go test ./hack/release/` all clean.

jveski (approved):

PR #655, opened 2026-08-24
relctl: one implementation of the versioning rule, shared by humans and CI

plombardi89:
## Problem

Release state has no single place to look. `RELEASING.md` said so itself: *"There is no single dashboard."* Whether `main` is releasable, which candidate trains are live, whether a release will be marked Latest, where one has got to across three workflows - each of those needed a different `gh` invocation, or could not be answered at all.

The versioning rule deciding all of it lived in ~500 lines of shell that only CI ever called. Building a CLI on top of that would have put the rule in two places, and the two would drift.

## Solution

The rule moves into Go and the pipelines call that, so humans and CI share one implementation instead of agreeing by convention.

`relctl` is the result. `status`, `next`, `preflight`, `classify` and `watch` answer questions about a release; `cut`, `rc`, `promote`, `branch create`, `soak` and `publish` drive one. The shell it replaces is deleted rather than left alongside.

`gh` stays authoritative: the workflows are the interface and `relctl` is a convenience over them. The one exception is version resolution, where `release-prepare` invokes `relctl next` itself, so the tool and the pipeline agree by construction rather than by review.

`RELEASING.md` is reorganised around the two questions people actually arrive with - how to cut a minor or major, and how to cut a patch for a series that already shipped - with the reference material behind them.


hack/cmd/relctl/relctl/version/resolve_test.go
```diff
@@ -0,0 +1,218 @@
...
+// commit first, so a fixture can express a train whose candidates are behind
+// HEAD, which is the shape promote has to get right. `<tag>@off` places the tag
+// on a commit that is NOT an ancestor of HEAD, as a tag cut on someone else's
+// branch would be.
+func fixture(t *testing.T, tags []string) string {
+	t.Helper()
+
+	dir := t.TempDir()
+
+	git(t, dir, "init", "-q")
```
plombardi89: `fixture()` inherits the ambient git configuration. It sets `user.email` and `user.name` locally, but nothing else.

`hack/release/next_version_test.go` already solves this for the shell suite, with a comment naming the reason:

```go
// The fixtures create throwaway repositories and tag them, so the ambient
// git configuration has to be neutralised: a maintainer with
// tag.gpgSign=true, or a commit template, or a hook path, would otherwise
// fail every case for reasons that have nothing to do with the resolver.
command.Env = append(os.Environ(),
    "GIT_CONFIG_GLOBAL=/dev/null",
    "GIT_CONFIG_SYSTEM=/dev/null",
    ...
```

The Go port dropped that guard. `commit.gpgsign`, `tag.gpgSign`, `core.hooksPath`, `commit.template` and `init.templateDir` all break it, and the failure mode is 76 cases here plus 345 in the differential going red at once with no obvious cause.

Setting `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` and the identity vars in the `git()` helper at line 110 covers every call site at once.

hack/cmd/relctl/relctl/version/differential_resolve_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	if _, err := os.Stat(script); err != nil {
+		t.Skipf("%s not present; the shell oracle has been removed", script)
+	}
+
+	cmd := exec.Command("bash", script) //nolint:gosec // fixed script path
+	cmd.Dir = dir
+
+	cmd.Env = append(os.Environ(), "MODE="+tc.mode)
```
plombardi89: `os.Environ()` leaks the developer's environment into the oracle.

`next-version.sh` reads `BUMP`, `SERIES`, `PRE`, `VERSION` and `ALLOW_CONCURRENT_TRAINS` with `:-` defaults. `MODE` is appended so it is always overridden, but the other five are only overridden when `tc.env` happens to contain them. Every bare `promote` case has `env: []`, and `TestResolveMatchesTheShellOnGeneratedTagSets` sets only `BUMP=`, so four inputs leak into all 207 of those runs.

Meanwhile `resolveCase.request()` builds the Go side from `tc.env` alone, so the two implementations are not seeing the same inputs. `VERSION` in particular is a plausible thing to have exported.

This does not make the port wrong. It makes "345 comparisons, all agreeing" a claim that holds on a clean environment, which CI is and a laptop may not be. Since that number is the load-bearing evidence in this PR, the shell should get an explicit minimal env rather than `os.Environ()`.

hack/cmd/relctl/relctl/version/resolve_test.go
```diff
@@ -0,0 +1,218 @@
...
+	for _, tc := range resolveCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			dir := fixture(t, tc.tags)
+			repo := NewGitRepo(t.Context(), dir)
+
+			result, err := Resolve(repo, tc.request())
+
+			if tc.want == "ERROR" {
```
plombardi89: 34 of the 76 cases stop here, asserting only that `err != nil`.

Breakdown of `resolveCases`: 37 assert a tag, 5 assert a base, 34 assert nothing but non-nil error. That is 45% of the suite that has to protect this code once the shell is gone.

"second train refused by default", "tag outside the series refused" and "promote rejects a stale train by name" are three distinct guards over three distinct number-space failures, and nothing here tells them apart. A refactor that collapsed every refusal into one would pass all 34.

`branch_test.go:64` already does the right thing with a substring match. These cases should too.

The honest cost: the fixtures were machine-generated from the shell's own `expect` invocations, which only recorded ERROR, so the generator cannot supply the expected messages. This is hand work. It is also the single highest-value thing left, because it is exactly what the shell's deletion is trading against.

hack/cmd/relctl/relctl/version/resolve.go
```diff
@@ -0,0 +1,471 @@
...
+	// Pre is an explicit prerelease suffix, e.g. rc.3.
+	Pre string
+	// Version is an explicit final for promote.
+	Version string
+	// AllowConcurrentTrains permits starting a second live train.
+	AllowConcurrentTrains bool
+}
+
+// Result is the resolution, plus everything the operator needs to see.
+type Result struct {
```
plombardi89: Every field but `Tag` and `Base` is unasserted, in both suites. `TestResolve` checks the tag or the base and returns; the differential compares those same two fields and nothing else.

So `LatestFinal`, `Live`, `Stale`, `Report` and `Warnings` are verified only indirectly, through whichever tag happens to fall out of them.

`discover()` is the code that fixes the failure this whole design exists to prevent. v0.1.24 reached rc.18 and was orphaned because promote guessed between trains, and the live/stale split is the answer to that. It is also the model the rest of the CLI is going to be built on, since "which trains are live" is a read command. It should be asserted directly.

hack/cmd/relctl/relctl/version/resolve.go
```diff
@@ -0,0 +1,471 @@
...
+
+// zeroVersion is the floor used when a repository has no finals yet.
+const zeroVersion = "v0.0.0"
+
+// Repo is the git history a resolution is computed against.
+//
+// An interface because every question here is about tags and ancestry, and
+// answering them from a fixture is what makes the resolver testable without a
+// repository per case.
+type Repo interface {
```
plombardi89: > An interface because every question here is about tags and ancestry, and answering them from a fixture is what makes the resolver testable without a repository per case.

Every test builds a repository per case. 76 in `TestResolve`, 76 more in the differential, 207 more in the generated run, each with a `git init` plus several commits, plus a bash process for the 283 differential cases. All of it on every local `make test`, which also implies `lint`.

The differential genuinely needs real git, because it has to feed the shell the same fixture. `resolve_test.go` does not.

A fake `Repo` for the permanent suite would be fast and deterministic, remove the ambient-config exposure entirely, and let you assert `Live`/`Stale`/`Report` precisely. That is one change closing three of the comments on this review.

hack/cmd/relctl/relctl/version/git.go
```diff
@@ -0,0 +1,115 @@
...
+// A tag whose commit later leaves the branch's history stops being seen. That
+// fails safe, since the resolver would recompute a version whose tag already
+// exists and refuse it, and it is the correct reading anyway: a tag on a commit
+// that is no longer on the branch is not part of its history.
+func (g *GitRepo) ReachableTags(pattern string) ([]string, error) {
+	// A repository with no matching tags is not an error, and neither is one
+	// with no commits at all, so failures here are reported as empty.
+	out, err := g.run("tag", "--merged", "HEAD", "--list", pattern)
+	if err != nil {
+		return nil, nil //nolint:nilerr // absence is not failure; see above
```
plombardi89: The comment's justification does not hold, and I checked rather than assumed.

`git tag --merged HEAD --list 'v[0-9]*'` exits **0** with empty output when nothing matches. It exits 128 only on an unborn HEAD, a non-repository, or a cancelled context. So "a repository with no matching tags is not an error" is already true without the swallow, and what this actually catches is real failure.

Today it is safe only because `Resolve` calls `repo.Head()` unconditionally before mode dispatch, and that does surface the error. That is an undocumented, untested invariant holding up a correctness property: if `Head()` ever moves or becomes conditional, a cancelled context silently becomes "no tags", `latest()` returns `v0.0.0`, and the resolver computes a version from a floor that is not real.

It is faithful to the shell's `|| true`, but Go can just return the error and let `Head()` report the unborn-HEAD case with a better message than "no tags".

hack/cmd/relctl/relctl/version/git.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return lines(out), nil
+}
+
+// TagExists reports whether a tag exists anywhere in the repository.
+func (g *GitRepo) TagExists(tag string) (bool, error) {
```
plombardi89: Same shape as `ReachableTags`, sharper consequence: `err` is discarded and a git failure is reported as "the tag does not exist".

That is the answer that lets `validate` fall through its existing-tag guard and proceed toward creating a duplicate. The ancestry check catches it in every path I traced, but by luck rather than by design, and the signature already returns an `error` that is never non-nil.

plombardi89 (commented):
Read the full diff against both shell originals, the workflow that still calls them, the NOTICE collector and the lint config.

**The port is faithful.** I walked every branch of `next-version.sh` against `resolve.go`/`modes.go` and found no divergence in the resolved tag or base. Report ordering matches, mode dispatch matches, the validation sequence (shape, series, tag-exists, ancestry, describe) matches, and both pieces of scar tissue survive intact: numeric `max_rc`, and repository-wide `has_final` kept distinct from reachability-scoped discovery.

The findings below are not in the port. They are in the evidence, and in what survives the oracle's deletion.

Three I would want settled before the shell goes:

1. **34 of the 76 permanent cases assert only that an error happened.** `branch_test.go` matches on the message; the resolve cases do not. That is 45% of the suite that would pass if every refusal collapsed into one.
2. **`Live`, `Stale`, `LatestFinal`, `Report` and `Warnings` have no assertions in either suite.** The differential compares `Tag` and `Base` only. The live/stale model is what the rest of the CLI is built on, and it is the exact thing the shell got wrong with the v0.1.24 orphan.
3. **The differential runs the shell with `os.Environ()`.** `next-version.sh` reads `BUMP`/`SERIES`/`PRE`/`VERSION`/`ALLOW_CONCURRENT_TRAINS` with `:-` defaults, and `request()` builds the Go side from `tc.env` only. "345 comparisons, all agreeing" holds on a clean environment; the generated 207 set only `BUMP=`.

A fake `Repo` in `resolve_test.go` is the cheapest route: it closes 1 and 2, removes the ambient-config exposure in the first inline comment, and cuts the local test loop substantially. The differential genuinely needs real git; the permanent suite does not.

The `gh` package landing ahead of its callers reads as deliberate (dependency approval settled early, command commits stay small). Noting it only so a later reviewer does not re-raise it.

**Nits not worth their own thread:** `relctl.go:32` `defer stop()` is unreachable before `os.Exit(1)`; the `strconv.Atoi` error branch in `prerelease` is dead since `preShape` already bounds to nine digits, and unlike `parseCore` it does not say why it is kept; the NOTICE commit message says it picks up "go-github and its go-querystring transitive", but `hack/cmd/notice/internal/gomod/gomod.go:35` is direct-deps-only and one entry was added, so the file is right and the message is not; `SplitRepo` defends the URL paste explicitly but accepts `Azure/un bounded`; `Options.BaseURL` is documented "for tests" with no test using it.

Worth considering putting the differential behind a build tag or `testing.Short()`. It is a temporary equivalence proof and it is currently paid for on every local `make test`, which also implies `lint`.

hack/cmd/relctl/relctl/version/branch.go
```diff
@@ -0,0 +1,84 @@
...
+// Every minor's patch space therefore belongs to exactly one branch, and main
+// never enters it. That is what makes a release branch possible at all: without
+// it, main and release-0.2 would both compute v0.2.5 and compete for the same
+// tag, which pre-1.0 they would do for months, since a minor series here runs
+// twenty-odd releases.
+//
+// major is never derived. A major is always a deliberate human decision, and
+// deriving one from commit contents would make it an accident waiting for the
+// right commit message.
+func ForBranch(branch string, major bool) (Policy, error) {
```
plombardi89: `bump-for-branch.sh:50-53` refuses when `MAJOR` is neither `true` nor `false`:

```bash
case "$MAJOR" in
  true | false) ;;
  *) fail "MAJOR must be true or false, got: ${MAJOR}" ;;
esac
```

A `bool` parameter cannot express that, so the validation is gone. Arguably the right narrowing, and it is not a bug today because nothing calls `ForBranch` yet.

Worth flagging because the differential only ever passes `MAJOR=%t`, so the corpus is structurally blind to the one input where the two implementations disagree. "0 skipped, all agreeing" is true and this case is simply not in it.

It becomes real in the commit that wires this to a workflow input, where the value arrives as a string.

hack/cmd/relctl/relctl/version/modes.go
```diff
@@ -0,0 +1,315 @@
...
+		return "", err
+	}
+
+	// Cutting the version a live train is heading for is not a fork, it is that
+	// train being finalised the long way round, so only warn when they differ.
+	if len(r.live) > 0 {
+		if contains(r.live, tag) {
+			r.warn("%s is the version its candidates were building toward, but mode=release cuts it from HEAD; use mode=promote to ship the tree that was soaked", tag)
+		} else {
+			r.warn("cutting %s while %s is still in flight; that train will be stranded",
```
plombardi89: The `::warning::` prefix is gone here, and `::error::` is gone from the failure path (`Run()` prints `error: %v`).

That is the right call for a library returning a `Warnings` slice, but it means whoever renders `Result` in the pipeline-switch commit has to re-add the Actions prefixes when running under CI. Miss it and "cutting vX.Y.Z while <train> is still in flight; that train will be stranded" silently stops appearing as a workflow annotation, which is the one place a human would have seen it.

Also a smaller loss nearby: `candidateCommit`'s error dropped the tag listing the shell included (`next-version.sh:239`, `(tags: $(git tag --list "${core}-*" ...))`). During an incident that listing is the useful half of the message.

hack/cmd/relctl/relctl/version/resolve.go
```diff
@@ -0,0 +1,471 @@
...
+			seen[name] = c
+		}
+	}
+
+	cores := make([]string, 0, len(seen))
+	for name := range seen {
+		cores = append(cores, name)
+	}
+
+	sort.Slice(cores, func(i, j int) bool {
```
plombardi89: Stdlib duplicates, on a module that is `go 1.26`:

- `sort.Slice` here, and the `sort` import, are `slices.SortFunc`
- `cmpInt` at line 158 is `cmp.Compare`
- `contains` at `modes.go:307` is `slices.Contains`
- `max` at line 421 shadows the builtin

None of these are caught by lint, and it is worth knowing why: `.golangci.yaml` never sets `linters.default`, so golangci-lint v2 runs only the standard five plus `goheader` and `wsl_v5`. The settings blocks for `varnamelen`, `godot`, `godox`, `nolintlint` and the rest are inert. `predeclared`, `containedctx`, `exhaustive` and `nilerr` are all off, which is also why the two `nilerr` threads above and the `switch tc.kind` with no `default` at `resolve_test.go:160` are human-review items rather than CI ones. Pre-existing, not this PR's problem, but it changes what a green build means here.

plombardi89:
All three settled, plus the nits. Verified each finding before acting on it.

## 1. 34 refusals asserting only that an error happened

Confirmed: **34 of 76, 44%**. `branch_test.go` matched on messages; the resolve cases did not.

Every refusal now names its reason. On its own that is not enough — per-case substrings would not notice the reasons converging — so `TestEveryRefusalNamesADistinctReason` pins the number of **distinct** reasons, currently 19 against a floor of 15. It is a floor, not a target.

## 2. No assertions on the train model

Confirmed: both suites asserted `Tag` and `Base` only.

`trains_test.go` now asserts `LatestFinal`, `Live`, `Stale`, `Report` and `Warnings` directly, including the v0.1.24 shape you named: candidates exist, no final was cut, a newer series shipped, so it must report as stale rather than live.

I also extended the **differential** to compare the state model, not just the Go suite. That turned out to matter more than expected — swapping live and stale does not change the computed tag, so the old proof could not have seen it. It now fails 77 cases.

## 3. The differential inheriting the environment

Confirmed, and it reproduces exactly as you describe:

```
$ SERIES=0.3 go test -run TestResolveMatchesTheShell
33 cases FAIL
```

Every variable the script reads is now set explicitly, defaulted to match `Request`, with only `PATH` and `HOME` inherited. Both sides see the same inputs by construction.

The proof:

```
clean env    ok
SERIES=0.3 BUMP=major PRE=rc.9 VERSION=v9.9.9 ALLOW_CONCURRENT_TRAINS=true MAJOR=true    ok
```

Since fixing input plumbing is exactly the change that can neuter a test, I re-confirmed it still bites afterwards: a minor bump that forgets to reset patch fails 31 cases.

## The fake Repo, and the gap it opened

Taking your suggestion closed 1 and 2 and dropped the permanent suite from 630ms to 6ms. It also created a hazard worth naming: with the permanent suite on a fake, and the differential — the only other code touching real git — deleted alongside the shell, nothing would check the fake against git. The suite would become a test of the fake.

`TestFakeRepoMatchesGit` closes that: same fixtures through both, requiring the same tag, latest final, live and stale, and the same base **identity**. Hashes cannot be compared, so the base is compared by what it points at, HEAD or the tag whose commit it is. Confirmed it bites by making the fake ignore reachability, which is the drift that would matter.

## Nits

All four taken. `defer stop()` was the real one: it sat above `os.Exit(1)`, which skips defers, so the signal handler leaked on every failing invocation. `Atoi` stays but now says why, matching `parseCore` — the silence was the complaint. `SplitRepo` validates the shape GitHub allows. `BaseURL` has an `httptest` case, which is what would notice if go-github changed how a base URL is applied.

**That work turned up something worse than any nit.** `relctl bogus-subcommand` printed help and exited **zero**: cobra skips argument validation entirely for a command it cannot run, so `Args: cobra.NoArgs` did nothing on a root with no `RunE`. A mistyped release command looking like success is the sort of fault nothing notices, because the tool appears to work having done nothing. Fixed, with tests.

## Not taken

**Gating the differential** behind a build tag or `testing.Short()`. It is 3.6s against 6ms, and it is deleted in a few commits — but making a correctness proof run less often is the wrong direction, and a build tag would need a CI change for something short-lived. Happy to revisit if it becomes annoying.

**The NOTICE commit message** was wrong, as you spotted: `gomod.go:35` is direct-deps-only and one entry was added. Amended and force-pushed with the author's sign-off, since this repo requires that for a rewrite.

You are right that the `gh` package landing ahead of its callers is deliberate — dependency approval was settled early so the command commits could stay small. Confirming so it does not get re-raised.

## State

501 assertions in `version/`, 0 skipped. `make lint` 0 issues, `CI=1 make test` green, `notice-check` up to date.

hack/cmd/relctl/relctl/version/classify.go
```diff
@@ -0,0 +1,196 @@
...
+	if match := seriesOf.FindStringSubmatch(tag); match != nil {
+		sameSeries, err := repo.AllTags("v" + match[1] + ".*")
+		if err != nil {
+			return err
+		}
+
+		candidates = append(candidates, sameSeries...)
+	}
+
+	highest := highestFinal(candidates)
```
plombardi89: **A git failure here marks the release Latest.**

`ReachableTags` and `AllTags` both return `nil, nil` on any git error. If both come back empty, `highestFinal` returns `""`, and the branch below sets `c.Latest = true` with "nothing has shipped, so nothing can outrank this".

The tool this absorbs refused that case on purpose. `hack/cmd/semver/main.go`:

```go
if len(tags) == 0 {
    // Distinguished from "no finals among the tags" below. An entirely
    // empty stream means the caller's `git tag` produced nothing, which is
    // a broken invocation rather than an answer, and answering "false"
    // would let a release deploy on the strength of a failed command.
    return errors.New("no tags on stdin; expected the output of `git tag --list`")
}
```

Two states, deliberately kept apart: an empty stream is a broken invocation, and tags-with-no-finals is a real answer. The port collapses them into one, and the merged state fails toward `Latest = true`.

That direction is the wrong one. In the resolver, swallowing a git error fails toward *refuse*, because ancestry catches it. Here it fails toward *ship*, and `Latest` is what `releases/latest/download` resolves to, which is the install command in README.md and every guide.

Not reachable through a total git failure, since `CommitOf` above returns a real error. Reachable through a partial one, including a cancelled context between that call and this one.

hack/cmd/relctl/relctl/version/git.go
```diff
@@ -0,0 +1,125 @@
...
+	out, err := g.run("tag", "--merged", "HEAD", "--list", pattern)
+	if err != nil {
+		return nil, nil //nolint:nilerr // absence is not failure; see above
+	}
+
+	return lines(out), nil
+}
+
+// AllTags lists tags matching a glob anywhere in the repository.
+func (g *GitRepo) AllTags(pattern string) ([]string, error) {
```
plombardi89: The swallow is inherited by reference (`see ReachableTags`), and the justification is weaker here than it was there.

I checked both. `git tag --merged HEAD --list <glob>` exits 128 only on an unborn HEAD. `git tag --list <glob>` does not even do that: I ran it in a freshly `git init`ed directory with no commits and it exits **0** with empty output. So there is no legitimate absence case for this one to be absorbing at all, and what it catches is a genuine failure.

That would be a small thing on its own. It is not, because unlike `ReachableTags` this feeds the `Latest` decision, where empty means "nothing outranks this". See the thread on `classify.go`.

hack/cmd/relctl/relctl/cmd_watch.go
```diff
@@ -0,0 +1,209 @@
...
+		result.Release = release.State()
+
+		// Published is the end of the line. A draft is not: the soak may still
+		// be running, and it is the soak that publishes.
+		result.Done = !release.Draft
+	}
+
+	// A failed build or soak ends it too, and reporting "still going" while
+	// nothing is running would wait out the timeout for no reason.
+	if progress.Build != nil && progress.Build.Done() && !progress.Build.Succeeded() {
```
plombardi89: **A failed soak never ends the watch.**

This check exists for the build, with the reason stated right above it: "reporting 'still going' while nothing is running would wait out the timeout for no reason." There is no equivalent for `progress.Soaks`.

A failed soak leaves the release as a draft. `Done` is set from `!release.Draft`, so it stays false, and the loop polls every 20 seconds until the 90 minute timeout and then returns "gave up waiting for %s". Nothing is running for almost all of it.

`watchVerdict` already knows the right answer and says so: "a draft already implies it did not [pass]". It just never gets to run. The newest soak is `progress.Soaks[len-1]`, and if it is `Done()` and not `Succeeded()`, nothing further will happen without a manual dispatch.

hack/cmd/relctl/relctl/relctl.go
```diff
@@ -0,0 +1,110 @@
...
+type Options struct {
+	// Repo is the owner/name slug commands talk to.
+	Repo string
+	// RepoPath is the local clone version resolution reads.
+	//
+	// Exists because release-prepare runs relctl built from main against a
+	// RELEASE BRANCH's history: the branch supplies the history, main supplies
+	// the tooling that computes against it. Getting this wrong resolves the
+	// wrong series and is the sort of thing that only shows up at release time.
+	RepoPath string
```
plombardi89: **`Options` has no seam for a stub, and three of the four new commands cannot be tested without one.**

`Options` carries `Repo`, `RepoPath` and `Output`. `Options.client` calls `gh.New(ctx, gh.Options{Repo: o.Repo})` and never sets `BaseURL`, so there is no way for a `relctl`-package test to point `status`, `preflight` or `watch` at an `httptest` server.

That looks like the reason for the coverage split rather than a coincidence: `next` is the one command needing no client, and `next` is the one command with tests. `cmd_status.go`, `cmd_preflight.go`, `cmd_watch.go` and `gh/releases.go` come to 767 lines with no test file between them.

`gh.Options.BaseURL` just gained a test in the neighbouring package, and `gh/runs_test.go` already has a stub that filters `head_sha` the way GitHub does. Threading a `BaseURL` (or a client factory field) through `Options` makes all of that reachable.

Two pieces need no seam at all and could be tested today: `judgeNightly` is a pure function over `[]gh.Run` that encodes a release gate (a red nightly blocks, a green one older than 48h blocks), and `watchVerdict` is a pure function over `watchResult`.

hack/cmd/relctl/relctl/cmd_next.go
```diff
@@ -0,0 +1,163 @@
...
+		Live:        result.Live,
+		Stale:       result.Stale,
+		Warnings:    result.Warnings,
+	}
+
+	switch opts.Output {
+	case OutputJSON:
+		return writeJSON(out, answer)
+
+	case OutputGitHub:
```
plombardi89: `--branch` is policy only, and is never checked against what is actually checked out.

`ForBranch(branch, major)` decides the bump and series, and `version.Resolve` then reads tags by reachability from local HEAD. The two inputs are independent, and `--branch` defaults to `main` unconditionally.

So on a `release-0.3` checkout, the default `relctl next` applies main's minor-bump policy to that branch's history and answers confidently. In the workflow this is safe, because release-prepare checks out the branch it was given. At a terminal it is the ordinary case for a maintainer patching a release.

`TestNextAppliesTheBranchPolicy` currently encodes the decoupling as intended: the `--branch release-0.4` case runs against `tagRepo`, whose HEAD is not that branch and only passes because `v0.4.0` happens to be in series 0.4.

Cheapest fix is probably to default `--branch` from `git rev-parse --abbrev-ref HEAD` rather than to `main`, and to refuse rather than warn when an explicit `--branch` disagrees with the checkout.

Separately on this output format: it emits `tag=` and `base=` only, but `release-prepare.yaml` reads `steps.bump.outputs.bump` in *Push tag* and `steps.bump.outputs.series` in *Compute tag*. Deleting `bump-for-branch.sh` needs both, and `nextResult` already carries them.

hack/cmd/relctl/relctl/cmd_status.go
```diff
@@ -0,0 +1,205 @@
...
+
+	result := statusResult{}
+
+	// Local first, so the version half still reports when the API is
+	// unreachable or uncredentialed.
+	resolved, err := version.Resolve(opts.repo(ctx), version.Request{
+		Mode: version.ModeRelease,
+		Bump: version.BumpMinor,
+	})
+	if err == nil {
```
plombardi89: A local resolution failure is swallowed and rendered as a fact.

If `version.Resolve` fails, `LatestFinalTag`, `NextFromMain`, `Live` and `Stale` all stay at their zero values, and `writeStatusText` prints `Live trains:  (none)` and drops the stale row entirely. "I could not tell" and "there are none" are different answers, and for the one command whose entire purpose is being the dashboard, the second is the wrong one to guess.

The comment above reads "Local first, so the version half still reports when the API is unreachable or uncredentialed", but the `if err == nil` guards the opposite direction: it is what happens when the *local* half fails. Ordinary causes are a stale checkout, a wrong `--repo-path`, or running outside a clone.

Same shape in `runPreflight`, where a failure silently drops the live-train note. Lower stakes there, since it is a note rather than a blocker, but the same swallow.

Also on this struct: `NextFromMain` is resolved against local HEAD, so on a release branch checkout it is "next from whatever is checked out" under a label that says main. Same root cause as the `--branch` thread on `cmd_next.go`.

hack/cmd/relctl/relctl/gh/releases.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return &out, nil
+}
+
+// Drafts lists unpublished releases, newest first.
+//
+// Worth surfacing because a draft is a release that built but never shipped,
+// and the usual cause is a soak that failed. They accumulate silently
+// otherwise.
+func (c *Client) Drafts(ctx context.Context, limit int) ([]Release, error) {
```
plombardi89: One page, filtered client-side, and the result is reported as a total.

`ListReleases` is called with `PerPage: limit` (default 30) and no pagination, then drafts are selected from what came back. `status` prints `"%d draft(s): built but not published"`, which reads as a count of all drafts.

The commit message says a real run turns up 24 drafts going back to v0.1.17. That is 24 out of a 30-release window, so the number is already close to the point where it silently becomes a lower bound, and it will cross it as more releases land.

`ReleaseBranches` in this same file paginates properly and is the model.

hack/cmd/relctl/relctl/gh/runs.go
```diff
@@ -0,0 +1,203 @@
...
+	HeadSHA string
+	// Limit caps how many are returned. Zero means 20.
+	Limit int
+}
+
+// Runs lists workflow runs, newest first.
+func (c *Client) Runs(ctx context.Context, q ListRuns) ([]Run, error) {
+	limit := q.Limit
+	if limit == 0 {
+		limit = 20
```
plombardi89: `Limit` is documented as "caps how many are returned" but is wired to `PerPage`, with no pagination. GitHub caps `per_page` at 100, so `Limit: 200` silently becomes 100 rather than fetching 200. Nothing passes more than 50 today, which is why it does not bite yet.

Related, in `cmd_status.go`: `inFlight` fetches 10 runs per workflow and then filters `!run.Done()` in Go. `ListWorkflowRunsOptions` has a `Status` field, so `in_progress` and `queued` can be filtered server-side and exactly. As written, a workflow with ten completed runs newer than a still-running one drops it from the dashboard. Concurrent tag pushes are the way that happens.

The `Progress` correlation itself looks right, and `runs_test.go` covers the part that is genuinely hard. One edge the design leans on and the tests do not reach: `until` only closes the window on a sibling strictly `After(build.CreatedAt)`. GitHub's `created_at` has second resolution, so two builds of the same commit created within one second leave the window open-ended and merge both builds' soaks into one report.

hack/cmd/relctl/relctl/version/resolve_cases_test.go
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package version
+
+// resolveCases are the fixtures and expectations ported from
+// hack/release/next-version-test.sh.
+//
+// Generated once by having bash itself parse that suite"s expect
```
plombardi89: `suite"s` should be `suite's`. The scripted edit that added `wantErr` mangled the apostrophe.

plombardi89 (commented):
Second pass, over `e14bb31e`..`1da4c2a6`.

## The previous round was addressed properly

| Prior finding | Status |
|---|---|
| Error identity | Fixed. All 34 refusals carry `wantErr`, plus `TestEveryRefusalNamesADistinctReason` as a meta-guard. That guard is a better idea than what I asked for: per-case substrings alone would not notice the reasons converging. |
| `Live`/`Stale`/`Report` untested | Fixed. `trains_test.go` asserts the model directly, and the differential now parses the shell's stderr report and compares it too. "Swapping live and stale does not change the computed tag" is the right observation, and 77 cases now fail if you do. |
| Env leak in the differential | Fixed, and verified rather than assumed. "Exporting SERIES=0.3 made 33 cases disagree" establishes the bug was real, not theoretical. |
| `Repo` interface unused | Fixed, and better than asked. The fake landed *with* `TestFakeRepoMatchesGit`, comparing base by identity rather than by hash. I did not ask for the conformance test; it is the thing that makes the fake safe to trust, and mutating reachability out of it to prove it bites is the right check. |
| Nits | `defer stop()`, the `Atoi` branch, `SplitRepo`, `BaseURL` all handled. The cobra `NoArgs` bug found along the way is a worse fault than anything I raised: a mistyped release command exiting zero. |

One prior finding is only partly closed, and two are now load-bearing in a place they were not before. Both are inline.

## New findings

The two that matter are `classify.go` failing open toward `Latest = true` on a git error, and `watch` never ending on a failed soak. Both inline.

Behind those, a coverage question. `cmd_status.go`, `cmd_preflight.go`, `cmd_watch.go` and `gh/releases.go` are 767 lines with no test files, and `relctl.Options` has no seam to add them through. Given this PR spent three commits making the resolver's evidence hold up, the command layer arriving untested stands out.

## Nits

`runSummary.Ref` holds a head branch for builds and an event name for soaks under one JSON key, so a consumer sees `"ref": "workflow_run"`. `watchVerdict` compares `State != "success"` stringly when `Run.Succeeded()` exists, and `State()` returns `"completed"` for a done run with no conclusion, which would render as a build failure. `runNext` takes both `out` and `cmd` when `out` is `cmd.OutOrStdout()`. `sort.Slice` appears twice more in `gh/runs.go`. `baseLabel` hand-rolls `strings.TrimSuffix` to avoid importing `strings`.

Hermeticity is still open, though narrower than before: `TestResolve` moved off git, but `fixture` (`resolve_test.go:84`), `classifyFixture.build` and `tagRepo` are three copies of the same builder and none neutralises ambient git config. `git_test.go` and both classify suites are permanent and still touch real git, so `commit.gpgsign=true` still fails them. `hack/release/next_version_test.go` remains the model.

## One question for the pipeline switch

`next --output=github` emits `tag=` and `base=` only, and `TestNextGitHubOutputMatchesTheWorkflowContract` pins exactly two lines. But `release-prepare.yaml` reads `steps.bump.outputs.bump` in its *Push tag* step and `steps.bump.outputs.series` in *Compute tag*. Deleting `bump-for-branch.sh` needs both keys, and `next` already computes them and exposes them in the JSON output. Worth deciding now whether `next` replaces both scripts or only `next-version.sh`, since the two-line assertion will resist the answer.

hack/cmd/relctl/relctl/version/resolve_test.go
```diff
@@ -0,0 +1,276 @@
...
+			continue
+		}
+
+		reasons[tc.wantErr]++
+	}
+
+	// Chosen from the reasons the resolver actually distinguishes today. It is
+	// a floor, not a target: raise it when a refusal gains its own message,
+	// and be suspicious of anything that lowers it.
+	const minimumDistinct = 15
```
plombardi89: The floor is 15, but the table currently names 19 distinct reasons.

So the guard tolerates four of them collapsing into an existing message, which is the degradation it was written to catch. The comment already argues the right principle ("be suspicious of anything that lowers it"); the number just needs to match today's reality, or become an exact comparison against a named set.

Worth noting that two of the 19 are the same template with a different version interpolated (`no prerelease train found for v0.9.9` and `... v9.0.0`), so the semantic count is nearer 18. Either way it is above 15.

Separately, adding `wantErr` surfaced something worth a look: the case named `"existing tag refused"` now asserts `"is older than the latest final"`, because `promoteExplicit` rejects it before `validate` ever runs its tag-exists check. The name and the behaviour disagree. The shell suite had the same gap, so this is inherited rather than introduced, and `validate`'s tag-exists guard does have real coverage via `"off-branch rc name still refused"`. But the case is not testing what it says it is.

plombardi89:
All of it, across six commits. Verified each finding before acting on it.

## The two that mattered

**`classify` marked a release Latest when git failed.** You were right that it was reachable through a partial failure, and right about the direction being the problem. The part I'd add: the tool this port *absorbed* refused that case deliberately, with a comment saying why, and the port collapsed the two states. A change whose whole justification was fidelity lost a property the original had.

Fixed in `ae8754a8`. The tag being classified is itself a tag, so an empty candidate set cannot be honest; "tags exist but none is final" remains a real answer.

**`watch` never ended on a failed soak.** Fixed. The terminating conditions are now a pure function so they're testable without a client, and only the *newest* soak decides — an earlier failure followed by a running retry is a recovery in progress.

**The swallows are gone rather than narrowed.** I checked the exit codes independently and got your result: `git tag --merged HEAD --list` exits 0 on no match and 128 only on real failure; `git tag --list` exits 0 even with no commits. So there was nothing legitimate being absorbed. `TagExists` keeps the one genuine distinction — exit 1 is a missing tag, 128 is broken.

**Mutation testing found a gap on the third attempt.** Restoring the swallow in `GitRepo` changed nothing, because the failure tests use the *fake* `Repo`. Nothing exercised `GitRepo`'s own error handling at all. `TestGitRepoReportsQueryFailures` closes it.

## Hermeticity

Reproduced before fixing:

```
$ GIT_CONFIG_GLOBAL=<gpgsign=true> go test ./...
fatal: failed to write commit object
```

One builder now (`f7ebe15d`), with identity supplied through the environment so a fixture needs no `git config` at all. Verified against a config setting `commit.gpgsign`, `tag.gpgSign`, `core.hooksPath` and `init.templateDir` together.

That fix then surfaced something else: with the global config neutralised, `init.defaultBranch` is unset and git creates `master`. **Every fixture was a branch the tool refuses.** Now explicitly `main`.

## Coverage

`053a4e21` adds the `--base-url` seam, hidden because it's a test seam and not a supported route to GHES. Eleven cases now cover `status`, `preflight` and `watch`, including the two encoding release policy: a red nightly blocks, and a green one older than 48 hours blocks.

You were right that the seam's absence was the cause rather than a coincidence.

## The gh layer

All three under-reporting findings fixed in `f6d287d2`: `Drafts` paginates, `inFlight` filters `Status` server-side, `Limit` is named and capped for what it is. Plus the second-resolution edge you flagged — run IDs are monotonic and now break the tie. Confirmed that case fails without it.

## `--branch`

Now defaults to the checked-out branch and warns on an explicit mismatch. Worth noting what that exposed: on this very branch, `relctl next` used to answer `v0.5.0` as though from main. It now refuses, which is correct.

## The pipeline question

I checked what those outputs are for. `bump` feeds two report lines in *Push tag*; `series` only ever fed `next-version.sh`. So `next --output=github` emits all four keys, the separate step disappears, and the contract test is four lines rather than two.

## The two the port dropped

`::warning::`/`::error::` are added back at the point of rendering when `GITHUB_ACTIONS` is set — a library returning a `Warnings` slice shouldn't embed them, but losing the annotation is silent. And `candidateCommit`'s tag listing is restored; during an incident it's the useful half.

## Not fixed here

The `.golangci.yaml` finding is real and repo-wide — filed as #657 rather than fixed here, since enabling `linters.default` would surface findings across the whole tree. I'd single out `nilerr`: it would have caught both `return nil, nil` bugs above, including the one that marked a release Latest.

Also inherited rather than introduced: `"existing tag refused"` still doesn't test what it names, since `promoteExplicit` rejects before `validate` runs. `"off-branch rc name still refused"` covers the tag-exists guard, so it's a naming problem rather than a hole. Left alone rather than renamed mid-review; say if you'd rather it changed.

## State

`CI=1 make test` green, `make lint` 0 issues, `notice-check` clean. Both hostile-environment checks pass: a git config with signing and hooks forced on, and `SERIES`/`BUMP`/`PRE`/`VERSION`/`ALLOW_CONCURRENT_TRAINS`/`MAJOR` all exported.

RELEASING.md
```diff
@@ -29,123 +29,413 @@ trigger.
...
+```sh
+gh workflow run release-prepare.yaml --repo Azure/unbounded \
+  -f mode=release -f branch=release-0.3 -f dry_run=true
 gh workflow run release-prepare.yaml --repo Azure/unbounded \
   -f mode=release -f branch=release-0.3
 ```
 
+</details>
+
+Step 2 is worth doing rather than skipping: `relctl next` resolves against the
```
plombardi89: This paragraph is exactly right, and it is the reason step 4 at `:250` is a problem.

Step 2 has the reader run `git checkout release-0.3` at `:244`. Step 4 is then:

```sh
relctl classify v0.3.1     # needs a main checkout; see the command's help
```

`Classify` derives `FromMain` from `IsAncestor(commitOf(tag), HEAD)`. On `release-0.3` the tag is reachable, so it reports `FromMain: true`, which is the opposite of correct, and computes `Latest` from that branch's trunk rather than main's. Both answers are wrong and neither errors. The trailing comment is the only guard, and it arrives after the reader has been told to check the branch out.

**The ordering is the symptom; the guard is missing from the command.** You wrote this warning for `next` and built it into the tool in `a3d34bf9`: default `--branch` from the checkout, warn when an explicit one disagrees. `version.Classify`'s doc comment says "Expects repo to be a checkout of the DEFAULT BRANCH", and the command's `Long` repeats it, but `runClassify` never checks. `classify` is the one command left where a stated precondition is enforced by nothing.

Comparing `CurrentBranch()` against the default branch and warning would close it, and unlike reordering these four steps it also protects the case where someone runs `classify` outside this procedure.

The stakes are the ones `d94c7c17` established: `from_main` decides whether `release-upgrade` skips deploy, Orca and smoke and publishes anyway.

RELEASING.md
```diff
@@ -29,123 +29,413 @@ trigger.
...
+  nowhere, and it does not sit as a draft waiting for anyone.
+  [Why](#release-branches-do-not-soak)
+- **It may not be marked Latest.** A patch on an older series leaves the install
+  command pointing at the newer release, so shipping `v0.3.1` after `v0.5.0`
+  changes nothing for new users. [The rules](#what-gets-marked-latest)
+- **A branch cut from an old tag runs no CI at all**, because the trigger lists
+  live in the branch's own copy of `.github/workflows/`, and the `release-*`
+  ruleset requires those checks - so every pull request to it would be
+  unmergeable. You will be warned rather than having to remember:
+  `relctl branch create 0.3 --dry-run` resolves the branch point and says so
```
plombardi89: `relctl branch create 0.3 --dry-run` does not resolve the branch point, and does not warn about the CI triggers.

`branchCommand` in `cmd_cut.go` builds a static summary:

```go
Summary: fmt.Sprintf("This will open release-%s at the newest release in that series.", args[0]),
```

and `--dry-run` returns from `dispatch.run` before the client is built, so nothing is resolved and nothing is fetched. The branch point is resolved by the *workflow*, which is why the `gh` block at `:228` annotates `-f dry_run=true` with "resolve the branch point only" and has to actually dispatch to get it.

The two share a name and are different operations:

- `relctl --dry-run`: describe what would be sent, locally, send nothing
- workflow `dry_run=true`: dispatch, let the workflow resolve and report, create nothing

The document already draws this distinction, and draws it well, at `:135`: "the dispatch remains the only way to prove the WORKFLOW agrees, which is what makes it worth keeping before a first release from a branch nobody has cut from." This bullet contradicts it.

"You will be warned rather than having to remember" is the part most worth fixing, since it tells a reader they can stop remembering something they still have to remember.

Same claim again at `:220`: `# --dry-run to see the branch point first`. If the intent is to see the branch point, both should point at the `gh` form below, which is the one that produces it.

RELEASING.md
```diff
@@ -29,123 +29,413 @@ trigger.
...
 
-# 2. Cherry-pick the fix onto it through a pull request. Fixes land on main
-#    first and are cherry-picked down; never the other way round.
+# 2. Check what it will cut. Run this ON the branch, or pass --branch: the
+#    version is resolved against whatever is checked out.
+git checkout release-0.3 && git fetch --tags
+relctl next
 
-# 3. Cut the patch.
+# 3. Cut the patch. --branch defaults to the checkout, so on the branch this
```
plombardi89: The comment and the command under it disagree. This says `--branch` defaults to the checkout "so on the branch this is just `relctl cut`", and `:247` is `relctl cut --branch release-0.3`.

Both are correct behaviour, so it is only a question of which the example should show. Step 2 has just checked the branch out, so the bare `relctl cut` demonstrates the default the comment is pointing at, while passing the flag demonstrates the opposite of what the sentence says.

RELEASING.md
```diff
@@ -29,123 +29,413 @@ trigger.
...
+relctl next
+git log --oneline "$(relctl next -o json | jq -r .latestFinal)..origin/main"
+
+# 2. Cut it. Shows the version and every workflow input, then asks.
+#    Add --dry-run to see that without dispatching.
+relctl cut
+
+# 3. Follow it through build, soak and publish. Exits non-zero if it does not
+#    publish, so this can be the last line of a script.
+relctl watch v0.5.0
```
plombardi89: The two halves of this procedure use different versions: `relctl watch v0.5.0` here, and `gh release view v0.4.0` at `:132`, under a summary that calls it "the same thing with `gh` only".

`v0.5.0` looks like the intended one, since the table at `:471` treats `v0.4.0` as current and `v0.3.0 to v0.4.0` as the last minor, so the `gh` line is probably the one to change. `v0.4.0` is used consistently everywhere else, including Break glass and Recovery.

RELEASING.md
```diff
@@ -29,123 +29,413 @@ trigger.
...
-sanctioned overrides.
+The release tooling itself always comes from `main`, even when cutting from a
+release branch. The two supply different things: the branch supplies the history
+the version is computed against, which is what scopes resolution to its series,
+while `main` supplies the scripts that do the computing. An old branch would
+otherwise cut releases with whatever tooling existed when it was created.
+
+A consequence is that a change to the resolver takes effect only once merged to
+`main`. It cannot be rehearsed by dispatching this workflow from a branch,
+`dry_run` included; the Go tests under `hack/cmd/relctl` are how you test it
```
plombardi89: This paragraph was updated for the port, and reads correctly. The one directly above it was not: `:350` still says "`main` supplies the **scripts** that do the computing."

`498f3986` changed that exact sentence in the workflow's own comments, from "main supplies the scripts that compute it" to "main supplies the tool that computes it", and "would be executing scripts from a branch" to "code from a branch". After `28e5556f` deleted the shell, `:350` is the last place in the repository describing the release tooling as scripts.

One word, and the rest of the paragraph is still accurate.

plombardi89 (commented):
Doc-only pass over `RELEASING.md` at `e2a1d564`, checking every `relctl` claim against the code rather than reading it for prose. The guide-then-reference split works, and putting the `gh` form in a `<details>` next to each `relctl` form reads better than two separate sections would have.

Eleven issues. Five are inline; the rest sit on lines this PR did not touch, so they are below with citations.

## Not anchorable, but the same class

**`RELEASING.md:664` - "each takes a typed confirmation rather than accepting `--yes`" is true for one of the two overrides.**

`publish` has no `--yes` at all and requires the phrase. `relctl soak v0.4.0 --max-notready-nodes 7`, the very next example at `:672`, takes an ordinary `Continue? [y/N]` and accepts `--yes`. In `cmd_soak.go` the phrase is set only under `forceInit`.

That looks deliberate rather than accidental, and `39a557ad` argues for it: "An ordinary soak retry is not break-glass and is not made to feel like one." But raising the NotReady ceiling is listed here *as* break-glass, so the doc and the code disagree about which paths are break-glass rather than merely about wording. Worth settling in one direction.

**`--force-init` appears nowhere in this document.** It is a typed-confirmation path in `cmd_soak.go`, a declared `release-upgrade` input, and it runs `site init` against the soak cluster instead of `upgrade-apply`. Break glass reads as the register of sanctioned overrides ("Two sanctioned overrides"), and this is the third, absent from it. It is also the one the code treats as most dangerous: it is what escalates `soak` from `y/N` to a typed phrase. Either it belongs beside the other two, or the section should say why it deliberately does not.

**`RELEASING.md:538` - `bump` is described as something that gets consulted**, in a document that says twice that it does not exist: `:145` "There is no `bump` input" and `:345` "The bump follows from that choice and is not an input." Confirmed against the workflow: `release-prepare.yaml` declares `mode`, `branch`, `major`, `pre`, `version`, `dry_run`, `allow_concurrent_trains`, and no `bump`. The property the sentence states is real and worth keeping; it just needs saying without naming an input the reader has twice been told is not there.

**`RELEASING.md:637-656` duplicates `:32-44` rather than continuing from it.** `:34` and `:639` are the same sentence, and the `make relctl-build` block at `:40` is repeated at `:653`. "Every procedure **below**" is also wrong at `:639`: that is line 639 of 764, and everything it refers to is above. The structure the doc sets up is right, and `:44` states it ("The full account ... is in [relctl and gh]"), so this section should open with the account. The genuinely new material is the authority exception and the README link, and both are currently preceded by two paragraphs already read.

## Duplication that has already drifted

- `:205` and `:464` are the same paragraph about the nightly not covering release branches, with `--branch release-0.3` in one and `--branch release-0.4` in the other. Nothing keeps them in step, and they have not stayed in step.
- `:79-84` and `:443-447` repeat the nightly and CI `gh` block verbatim, comments included, though `:88` already links to the reference section for what the checks mean. Plausibly deliberate so the walkthrough stands alone, so I have not called it a defect.

The doc elsewhere prefers a link to a repeat, which is why these stand out rather than being a general style point.

## Scope

All of this is prose against behaviour that is already correct, with one exception: the comment on `:265` asks for a guard in `cmd_classify.go`, because the doc problem there is a symptom rather than the cause.

plombardi89:
All eleven addressed, plus a twelfth the review did not catch. Two commits: `9ea81548` (the guard) and `b736a01b` (the document).

**The `--dry-run` claim at `:194` was the worst of these and it was mine.** I wrote that `relctl branch create --dry-run` resolves the branch point and warns, under "You will be warned rather than having to remember". It does neither: `branchCommand` builds a static summary and `dispatch.run` returns before the client is built. I had read the workflow's `CI_OK` check, confirmed the *workflow* warns, and then attributed that to relctl's `--dry-run` — conflating the two operations the document distinguishes at `:135`, in a paragraph I preserved. Telling a reader they can stop remembering something they still have to remember is worse than not mentioning it. Both places now say which of the two produces the answer.

**`:265` — guard added, not just reordering.** `runClassify` now refuses when `CurrentBranch()` names a branch other than `main`, and the procedure checks main out first. Detached HEAD warns rather than refuses: `CurrentBranch()` already reports empty rather than guessing, and refusing on "cannot tell" would break a legitimate checkout to defend a case that cannot be detected. Safe in CI, since `release-upgrade.yaml:54` checks out `ref: default_branch` and gets a real branch.

I deliberately did **not** give `next` the same treatment, and the reason is structural rather than stylistic. `next --branch release-0.3` from a main checkout already exits 1 with `computed tag v0.4.1 is outside series 0.3` and writes nothing to stdout — the series guard and the repository-wide tag-exists check mean a version computed from the wrong history refuses rather than misleads. `classify` has neither, and no internal consistency check could catch it. The commit message records this so the asymmetry does not read as an oversight later.

**Break glass is a table of three.** Your point that the doc and code disagreed about *which* paths are break-glass was the right framing. `force_init` now has a section, and the three sit in a table showing what each bypasses and what confirmation relctl asks for, so doc and code can be compared at a glance rather than by reading prose.

**A twelfth, surfaced by a question about whether `next` needed the same fix:** `:267` said `next` "warns when the two disagree". For the pairing named in that same sentence it refuses. The doc undersold its own protection and named the wrong mechanism, which invites a reader to press on through a warning that never appears.

**On the duplication.** `:205`/`:464` had already drifted to different branch numbers, and it happened inside the commit that introduced it. That was the accepted risk of inlining facts into the guides, and it did not survive one commit. The reference copy now drops the branch-specific example and links to the guide; the relctl/gh section keeps only the material that is new by that point.

Rest as described: `:245` showed `--branch` while its comment described the default, `:132` used a different version from its own procedure, `:350` was the last place in the repo calling the tooling "scripts", `:538` named a `bump` input the document twice denies.

Left alone deliberately: the repeated `gh` nightly block, which keeps the walkthrough standalone, as you allowed.

Verified by extracting every `relctl` invocation from the document and checking each command and flag against the built binary's help, plus the anchor check (28 links, all resolving). Three mutations on the guard: removing it fails three tests, refusing on detached HEAD fails one, warning instead of refusing fails two.

jwilder (approved):

PR #659, opened 2026-08-24
Protect the controller-owned WireGuard port annotation from node-agent changes

ipochi:
The node agent can update networking annotations on Nodes, but the WireGuard port is assigned and managed by the controller. The admission policy now prevents the node agent from adding, changing, or removing that annotation, while still allowing the controller to manage it.

The tests compile and evaluate the rendered admission expression, covering allowed and denied add, change, remove, and unchanged cases.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -358,6 +358,15 @@ notices:
     license:
       - name: MIT License
         link: https://github.com/golang-jwt/jwt/blob/v5.3.1/LICENSE
+  - dependency: github.com/google/cel-go
+    ecosystem: go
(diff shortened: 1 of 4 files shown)


ipochi:
Not sure how to work this around , passes locally

```
❯ make orca-inttest
go test -tags=integrationtest -race -count=1 -timeout 15m ./internal/orca/inttest/...
ok      github.com/Azure/unbounded/internal/orca/inttest        21.657s
```

cchildress (approved):

PR #687, opened 2026-08-29
Support Azure Container Linux as an agent host

plombardi89:
Adds support for running the unbounded agent on Azure Container Linux, an
immutable Flatcar-derived image with a read-only `/usr` and no package manager.
Validated end to end: an ACL node bootstrapped by the agent joins an AKS cluster
and reaches `Ready`, with working cross-site pod networking.

## Why it did not work before

ACL reports `ID=azurelinux` with a 3.x `VERSION_ID`, so it was silently
classified as mutable Azure Linux 3. It is a different OS: `rpm` is not merely
absent from `PATH`, it does not exist. Three things blocked bootstrap.

- The agent could not write its own files to `/usr/local`.
- `InstallPackages` aborted on any host without apt-get, tdnf or dnf.
- `systemd-nspawn` could not be installed at all.

## Approach

**Identity.** Detect ACL via `VARIANT_ID`, with `ID_LIKE=flatcar` as a
deliberate second signal so an image that later drops `VARIANT_ID` cannot
silently regress into being treated as Azure Linux 3. The resolved rootfs image
is unchanged: ACL shares the Azure Linux 3 kernel, and the nspawn rootfs runs on
the host kernel rather than supplying its own.

**Install prefix.** `HostPrefix` on the agent config and
`spec.agent.hostPrefix`, defaulting to `/usr/local` so hosts that do not set it
are byte-identical to before.

This is deliberately *declared, not detected*. Where the agent may write is a
property of the filesystem, not of the distribution, so keying on distro
identity would misclassify a hardened host with a read-only `/usr` and would
silently relocate files on any host whose os-release changed. A wrong guess is
also expensive: the lifecycle helper path is baked as an absolute path into the
nspawn drop-in and the config regeneration unit, and that regeneration unit
reruns on every container start. Preflight refuses with an actionable message
rather than guessing.

The value rides on the persisted applied config, which is already checksummed
and re-read by the systemd-invoked pre-start hook. That is what lets processes
systemd starts resolve the same paths the bootstrap used; an environment
variable could not, since it would have to be propagated into every generated
unit and would silently degrade to the default when it was not.

**Package detection.** Probe the capability each required package provides
(`systemd-container` to `systemd-nspawn`, `nftables` to `nft`) rather than the
package name, so any immutable host is covered without enumerating it.

**systemd-nspawn.** Deliver `systemd-container` as a systemd system extension,
which is the mechanism ACL itself uses to extend `/usr`; it already merges
`containerd` and `oem-azure` this way.

## Two constraints found the hard way

Both are encoded in code and documented, because neither is inferable from the
code alone and both cost real time to establish.

**The extension must ship as a squashfs `.raw`, not a directory.** A directory
extension is merged by bind-mounting it out of `/var/lib/extensions`, which
carries SELinux xattrs from the writable root, and a binary with that label
cannot acquire a D-Bus name under Enforcing. `systemd-machined` times out,
`systemd-nspawn` reports `Failed to register machine: Access denied`, and **no
AVC denials are logged at all**, because the relevant rules are `dontaudit`ed.
A squashfs image carries no SELinux xattrs, so merged files take the filesystem
default, which is exactly what ACL's own read-only `/usr` uses.

**`systemd-nspawn` links `libsystemd-shared` by full package release**
(`libsystemd-shared-255-33.azl3.so`), and `systemd-sysext` matches only on `ID`
and `SYSEXT_LEVEL`, so it will merge an extension whose binaries cannot load and
surface a loader error much later. The builder bundles the library so the
extension is not pinned to one systemd build, and a compatibility guard checks
provenance against the running systemd before the host is touched.

## Validation

Real AKS cluster, real Azure Container Linux VMSS node, SELinux Enforcing
throughout:

```
NAME                   STATUS   VERSION   OS-IMAGE                    CONTAINER-RUNTIME
aclv2-dc1-acl-000000   Ready    v1.34.3   Microsoft Azure Linux 3.0   containerd://2.1.8
```

- Refusal verified: without `--host-prefix` the install fails immediately with
  `Read-only file system` and an actionable message.
- Generated units carry the relocated path, with zero `/usr/local` references
  anywhere under `/etc/systemd/system`.
- Cross-site pod networking verified bidirectionally over WireGuard (~1.3ms),
  and DNS resolves from a pod on the ACL node.
- Survives reboot: the extension auto-merges at boot, the nspawn machine
  auto-starts, and the pre-start hook re-resolves paths correctly from the
  persisted config.

The live run found three bugs that unit tests could not, each fixed here: the
host-packages preflight blocked its own remediation, the install task fetched
the source before checking what was already installed, and `manual-bootstrap`
built two divergent `AgentSpec` values so `--host-prefix` reached the config but
not the install script.

## Known limitation

First bootstrap on ACL takes two passes. The extension is merged, then the host
must reboot before D-Bus will activate `org.freedesktop.machine1`, because dbus
loads its policy at startup and reloading it is itself denied. The agent detects
this immediately after merging and reports it, rather than continuing and
failing later at `machinectl enable: Access denied` with no indication that a
reboot is all that is required.

Installing the extension at provision time, via Ignition, would make this
single-pass. That is follow-up work.

## CI

`agent-e2e-kind` cannot boot an ACL host, since ACL publishes no QEMU-ready VHD,
so the ACL path is validated manually as above. The existing Ubuntu and Fedora
matrix entries are unaffected: every default is preserved, and regression tests
pin the pre-existing paths for both the host layout and the blue-green binary
layout.

## Review notes

The 10 commits are individually reviewable and each builds standalone. The two
worth reading first are `agent: make the host install prefix declarable`, which
carries the declaration-over-detection reasoning, and `images: add Azure
Container Linux nspawn sysext builder`, which carries the SELinux and soname
constraints.

Also included, found while working in these files and separable if preferred:
removal of the dead `AGENT_DISABLE_OCI_IMAGE` path, which logged "falling back
to debootstrap" for an implementation that does not exist, and a `--image` flag
for `forge site azure add-pool`, which is what makes provisioning an ACL host
possible in the first place.


plombardi89:
Superseded by #713.

Rebuilt on current `main` and scoped down. The short version: Azure Container Linux `3.0.20260818` now ships `systemd-container` in its base image, verified by booting it, so the system-extension apparatus this PR carries is no longer needed. Nine commits dropped, including the sysext builder, the OCI publisher, the release job and the two CRD fields.

What carried over is the host support that is still required - the declarable install prefix, ACL distro detection, the capability probe - plus the Ignition bootstrap variant and new e2e coverage.

The `agent config e2e` failure here is fixed in #713: preflight was probing `<prefix>/libexec` directly, which fails ENOENT on a directory the agent creates, and that also broke stock Ubuntu with LocalDNS enabled.

Leaving this open for now since its history is useful while reviewing #713.

Issue #698, opened 2026-09-02
releng: replace the user-owned release tag deploy key with a GitHub App so release runs are not attributed to a person


plombardi89:
## Symptom

Every `unbounded release` (`release.yaml`) run reports the same `actor` regardless
of who cut the release. Two runs from 2026-09-02:

| Release build | `actor` | `triggering_actor` | Who actually cut it |
| --- | --- | --- | --- |
| [33655638847](https://github.com/Azure/unbounded/actions/runs/33655638847) (`v0.5.0`) | `plombardi89` | `plombardi89` on attempt 1 | `cchildress`, via prepare run [33655577293](https://github.com/Azure/unbounded/actions/runs/33655577293) dispatched 36s earlier |
| [33667392561](https://github.com/Azure/unbounded/actions/runs/33667392561) (`v0.6.0`) | `plombardi89` | `plombardi89` | `cchildress`, via prepare run [33667308864](https://github.com/Azure/unbounded/actions/runs/33667308864) dispatched 49s earlier |

The `v0.6.0` case is the clean one: cchildress dispatched, and both `actor` and
`triggering_actor` came back `plombardi89` on attempt 1. cchildress appears
nowhere on the run.

This holds for every release since 2026-06-23. Correlating all 80 `release.yaml`
runs against all 52 `release-prepare` runs: every release build with a matching
prepare run reports `plombardi89`, including ones cut by `bcho`, `vpatelsj`,
`jwilder`, `jveski` and `cchildress`. The only builds with a truthful actor are
the handful with no prepare run at all (`v0.1.24-rc.11` through `rc.18`,
`v0.2.0-beta.4`, `v0.2.3-alpha.0` through `alpha.2`), which were tags pushed by
hand.

## Cause

`release-prepare` pushes the tag over SSH with `RELEASE_TAG_DEPLOY_KEY` rather
than `GITHUB_TOKEN`, because GitHub suppresses workflow triggers for tags pushed
with the default token:

https://github.com/Azure/unbounded/blob/main/.github/workflows/release-prepare.yaml#L110-L116

GitHub attributes a deploy-key push to the account that **registered the key**.
That key was added by an individual maintainer when 6c8d83d (#328) introduced
this mechanism, so the actor on every workflow-cut release is that person,
permanently.

`triggering_actor` cannot substitute: on a push it equals `actor`, and on a
re-run it names whoever pressed re-run. Neither field can name the cutter.

## Why this is worth fixing rather than just documenting

[...]

PR #699, opened 2026-09-02
relctl: report who actually cut a release

plombardi89:
## What

`relctl watch <tag>` gains a `Cut by:` row and `relctl status` a `BY` column,
naming who actually cut a release.

```
Tag:      v0.6.0
Cut by:   cchildress  (release-prepare https://github.com/Azure/unbounded/actions/runs/33667308864)
Build:    success  https://github.com/Azure/unbounded/actions/runs/33667392561
Soak:     success  (workflow_run)  https://github.com/Azure/unbounded/actions/runs/33670115829
Release:  published
```

## Why

`release-prepare` pushes the tag with a deploy key rather than `GITHUB_TOKEN`,
because GitHub suppresses workflow triggers for tags pushed with the default
token. GitHub credits a deploy-key push to whoever **registered the key**, so
every release build names that same person. `triggering_actor` is no better: on a
push it equals `actor`, and on a re-run it names whoever pressed re-run.

Checked across all 80 `release.yaml` runs: every build with a matching prepare
reports the key owner, including releases cut by `bcho`, `vpatelsj`, `jwilder`,
`jveski` and `cchildress`.

## How

Nothing in the API links a tag push to the run that made it, so this is
correlation rather than lookup: find the `release-prepare` run whose execution
window contains the push and report who dispatched it. `gh.Attribute` is pure and
separate from the fetching for that reason.

| Rendering | Means |
| --- | --- |
| `Cut by: <login>` with a prepare link | That person dispatched the prepare that pushed the tag |
| `Pushed by: <login>` | Tag pushed by hand, so the actor really is the pusher |
| `Cut by: unknown`, or `?` in `BY` | Could not be established |

What it rests on, all commented in the code:

- Windows compare `created_at`, never `run_started_at`. A re-run moves the latter
  and leaves the former at the moment of the push.
- The candidate list is fetched **after** the runs it explains. A non-match reads
  as "pushed by hand", so a list taken too early would name the key owner as the
  pusher.
- Containment is unambiguous only because `release-prepare` is serialized by a
  `concurrency` group.
- A prepare that did not succeed is not a candidate, since the push is the last
  thing that job does.

`BY` never falls back to the raw actor as a default. The raw value stays in
`-o json` as `actor`, beside the derived answer under `by`.

`watch` also stops ending a 90-minute poll on one bad response: 5xx, rate limits
and connection failures are retried, definite answers like a 404 still fail at
once, and `--once` never retries.

## Testing

Table and named tests in `gh/attribution_test.go`, including that no path can
report the key owner and that correlating the same fixture on `run_started_at`
finds nothing. Command-level tests assert the request **order**, which no
inspection of the output can show. Mutation-checked. Verified live: `v0.6.0` →
`cchildress`, `v0.4.0` → `plombardi89`, `v0.2.3-alpha.2` → a hand push by `bcho`,
`v0.1.5` → unknown.

## Notes

A workaround for the credential, not a fix. #698 tracks replacing the deploy key
with a GitHub App — though an App token names the app, so it makes `actor` honest
rather than correct.

The `.golangci.yaml` change stops `make fmt` rewriting `cancelled`, which is
GitHub's spelling for a run conclusion, to a value the API never sends.


cchildress (approved):

plombardi89:
@copilot merge is failing on license-check. We need to update the files with MIT headers to Apache 2 headers that are compliant with the rest of the project.

jwilder (approved):

PR #705, opened 2026-09-03
gantry: Deterministic Cold Seeding with Rotating Lease Chairs design

vpatelsj:


internal/gantry/chairs/manager.go
```diff
@@ -0,0 +1,765 @@
...
+	for _, chair := range snapshot.Chairs {
+		if chair.Occupied() {
+			occupied[chair.ID] = struct{}{}
+		}
+	}
+
+	for index := range Count {
+		id := ID(index)
+		if _, ok := occupied[id]; !ok {
+			empty = append(empty, id)
```
plombardi89: **[P1] Obsolete occupied Leases can permanently prevent recovery**

After node-pool replacement, suppose all 64 Leases remain occupied, their holders and successors are permanently absent, replacement agents have new identities, and every assignment is older than the immediately preceding epoch. Startup treats these chairs as occupied and never claims them, but `chairs.Rank`/`SelectableCount` exclude their old assignments.

Demand-driven reclamation cannot rescue this: readiness requires eight selectable chairs (`cmd/gantry/main.go:800-802`), and even direct resolver invocation returns before dispatch/reclamation when fewer than eight chairs rank (`coldstart/chair.go:128-140`). Advancing epochs only makes the assignments older. The operator preserves existing Leases with `OpCreateIfAbsent`. The deployment remains NotReady indefinitely without external state repair.

Please provide a startup recovery path that probes and reclaims obsolete occupied assignments without first requiring eight selectable chairs. This should not steal merely expired but responsive chairs. Add a recovery test with retained old assignments, new identities, and multiple epoch advances.

cmd/gantry/chaircall.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	ln, err := tls.Listen("tcp", addr, tlsCfg)
+	if err != nil {
+		return nil, fmt.Errorf("chaircall: listen %s: %w", addr, err)
+	}
+
+	srv := &http.Server{
+		Handler:           coord.NewChairHTTPHandler(local, logger),
+		ReadHeaderTimeout: 5 * time.Second,
```
plombardi89: **[P1] Incomplete HTTPS bodies can exhaust resources and block shutdown**

`ReadHeaderTimeout` does not bound body reads. A client can complete POST headers, send part of a declared body, and stop indefinitely. `serveChairHTTP` blocks in `io.ReadAll` before assignment validation (`coord/chairhttp.go:84`), retaining a connection, handler, and potentially almost 1 MiB of buffering per request. There is no global handler-admission bound or client-certificate requirement. The origin-pull semaphore cannot protect this earlier path.

The existing libp2p handler has a whole-stream deadline and inbound stream cap; this HTTP path bypasses both. Additionally, `main.go:415` defers `chairStop(context.Background())`, so unfinished handlers can prevent process exit outside the normal shutdown budget.

Please add body/handler deadlines and admission bounds, plus bounded shutdown with forced connection closure. Test a client that completes headers but never completes its body.

internal/gantry/coord/chairhttp.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	dialer := &net.Dialer{}
+
+	raw, err := dialer.DialContext(ctx, network, target)
+	if err != nil {
+		return nil, err
+	}
+
+	conn := tls.Client(raw, chaircall.ClientTLSConfig(pid))
```
plombardi89: **[P1] RPC timeouts do not bound this custom TLS dial**

Go HTTP transport detaches connection establishment from the waiting request cancellation/deadline (`context.WithoutCancel` in `Transport.getConn`). `http.Client.Timeout` therefore bounds the caller waiting for a response, not this custom `DialTLSContext` operation. The zero-timeout `net.Dialer` and `HandshakeContext` have no independently bounded context here.

A chair address that accepts TCP but never completes TLS makes the RPC time out while leaving its socket and dial goroutine alive. Repeated requests can accumulate unfinished connections; idle-pool limits do not bound them and `MaxConnsPerHost` is unset. The stalled endpoint need not possess the pinned key.

Please establish a timeout inside `dialPinned` covering TCP establishment and TLS negotiation, close on expiration, and test repeated calls to an accepting TCP listener that never performs TLS. Go implementation: https://github.com/golang/go/blob/go1.26.6/src/net/http/transport.go#L1516-L1543

cmd/gantry/main.go
```diff
@@ -831,52 +786,33 @@ func runAgent(args []string) error {
...
-			return "members informer not synced", false
-		}
-
-		if requireSelfAnnounce && !selfAnnounced.Load() {
-			// Production + dynamic bootstrap: peers cannot
-			// discover us until our pods/patch lands. Staying
-			// 503 until then makes the rolling deploy pause and
-			// surfaces an RBAC misconfiguration immediately.
-			return "members self-announce pending (check pods/patch RBAC)", false
+		if chairManager != nil && !chairManager.Ready() {
```
plombardi89: **[P1] Eight-chair readiness can deadlock a mixed-version rolling upgrade**

On an eight-node deployment, shipped `maxUnavailable: 50%` initially replaces four agents. Each new agent can hold only one chair (`chairs/manager.go:352-354`), and the base-version agents cannot claim chairs. The four new pods never become Ready. If the four surviving old pods stay healthy, the availability budget is exhausted and the DaemonSet cannot replace them to reach eight chairs.

Removing legacy RBAC does not reliably break this deadlock: base agents latch membership initialization and self-publication, and can stay Ready using cached state and surviving DHT connections. This finding is conditional on the old pods remaining available; unavailable old pods can permit further replacement.

Please make rollout readiness compatible with the first replacement batch. Add a base-to-head rolling-upgrade test with eight healthy old agents; fresh installation of eight new agents does not exercise this dependency.

internal/gantry/chairs/manager.go
```diff
@@ -0,0 +1,765 @@
...
+	epoch := m.CurrentEpoch()
+	if m.electionEpoch != epoch {
+		m.electionEpoch = epoch
+		m.claimRound = 0
+		m.knownFull = false
+		m.participating = false
+		m.selectionReady = false
+	}
+
+	if m.held != nil || m.reserved != nil || m.knownFull || m.claiming || (m.selectionReady && m.bootstrapReady && !m.participating) {
```
plombardi89: **[P1] A full snapshot suppresses retries after failed bootstrap**

`observe` sets `knownFull` from occupancy even when `Connect` returns zero (`manager.go:674-695`). This guard then suppresses observation regardless of `bootstrapReady`, and non-holder maintenance does not retry connections.

A restarting non-holder that reads 64 occupied chairs while all initial bootstrap dials fail can remain disconnected after the network recovers. Without another successful static or incoming connection, its routing table stays empty and readiness remains false until an epoch reset, potentially the remainder of six hours. Static bootstrap is one-shot; Gantry does not configure a periodic DHT bootstrap callback guaranteeing recovery from an empty table.

Please separate completed election from successful bootstrap: a full snapshot should suppress claims, not connection recovery. Add a full-snapshot non-holder test that restores connectivity after failed initial dials. The existing bootstrap-retry test exercises a holder instead.

internal/gantry/mirror/mirror.go
```diff
@@ -1421,53 +1506,107 @@ func (s peerAttemptSummary) allStaleOrFiltered() bool {
...
+	// by ResolveStall - 60s for a 1 GiB layer against the 30s response-header
+	// timeout fail-open containerd applies. Round 0 therefore routinely outlives
+	// the client, and anything flushed after it returns is already too late.
+	// Emit the metadata-only origin HEAD result up front so the client stays
+	// attached for the wait instead of bypassing Gantry to origin. Gated on
+	// re-discovery being enabled so the budget <= 0 path below keeps returning
+	// firstResult untouched and its terminal legs reachable. A failed HEAD leaves
+	// headers unflushed, which is the pre-existing behavior.
+	if s.peerRediscoverBudget > 0 {
+		_ = s.beginBusyResponse(ctx, w, d, kind, upstream, repo, stream, logger)
```
plombardi89: **[P1] Slow origin HEAD blocks healthy warm P2P before discovery**

With production live streaming and positive rediscovery budget, this synchronously calls `origin.Head` on every local digest miss before checking the DHT. It uses the original request context without a short metadata deadline; the origin HTTP client allows five minutes (`origin/origin.go:361-365`).

If the registry stalls its HEAD response, an otherwise healthy peer already holding the digest is never discovered or fetched before the client response-header deadline. This creates an origin-responsiveness dependency even for already-authorized warm P2P requests. An immediate HEAD error permits P2P to proceed; a stalled request serializes the entire path.

Please preserve an origin-independent warm path and bound metadata requests needed for cold waiting. Test a healthy peer with a blocking origin HEAD under the production live-stream-plus-rediscovery configuration.

internal/gantry/coldstart/chair.go
```diff
@@ -0,0 +1,629 @@
...
+		if recheck.trustedFailure {
+			return nil, ErrFailureShortCircuit
+		}
+
+		sawTransientFailure = sawTransientFailure || recheck.transientFailure
+
+		// A missing provider record is the normal state while the seed cohort is
+		// still pulling, so recruiting a backup would only duplicate origin work.
+		if len(recheck.accepted) > 0 && patience < maxStillPullingRounds {
+			patience++
```
plombardi89: **[P1] Healthy queued pulls are treated as stalled and duplicated on backups**

The new pump reports `STARTED` before obtaining an execution slot (`cmd/gantry/main.go:1778-1816`). Queued work is detached, has no queue deadline, and starts its origin budget only after dequeue. This resolver nevertheless abandons the cohort after four poll windows even if it continues returning `ALREADY_PULLING`.

Production mirror calls pass size zero, yielding 30-second blob windows. Around 120 seconds of legitimate queueing or a healthy large download therefore recruits another eight chairs while original jobs remain queued or running. Successive cohorts add origin pressure and worsen the backlog. Dequeued jobs also enter `runOriginPull` without rechecking whether content became locally available.

Please distinguish queued/progressing work from genuinely stalled work and align takeover with queue/pull budgets. Recheck local content before executing queued downloads. Add pump/resolver integration coverage for healthy work exceeding the patience limit, not just isolated admission and permanently stalled-cohort tests.

plombardi89 (commented):
Code and architecture review focused on severe instability, failure recovery, concurrency, and resource bounds. These nine findings were source-reviewed against head 30cc4c65 and base a5cfed07, including relevant tests and design documents; they are not executed reproductions. Existing CI is passing, but the inspected tests do not cover these combinations.

The main architectural concern is coupling distinct states: occupied versus usable chairs, completed election versus successful bootstrap, and accepted versus progressing downloads. The HTTPS migration also bypasses protections previously provided by the libp2p RPC boundary.

Please also reconcile the rotating-chair design document with this head: Rolling Interoperability still promises legacy Pod/Node RBAC and endpoint publication, while the implementation removes them; Cold Pull Path says accepting cohorts are retained, while the resolver escalates after a fixed patience limit.

Submitting as a comment-only review. No critical data-corruption defect or unsynchronized-memory data race was verified in this pass.

internal/gantry/coldstart/chair.go
```diff
@@ -0,0 +1,629 @@
...
+		}
+
+		if ctx.Err() != nil {
+			return nil, ErrExhausted
+		}
+
+		// The accepted pull can fail after replying STARTED. Re-query those
+		// chairs once so terminal origin failures short-circuit before the
+		// resolver activates an ordered backup cohort.
+		recheck := r.dispatchChairs(ctx, accepted, registry, repository, kind, []digest.Digest{d}, trustedFailureClasses)
```
plombardi89: **[P1] Failure rechecks restart work after its cooldown expires**

This re-query is another work-starting `please_pull`, not a status observation. Production unknown-size blob polling waits 30 seconds, while the initial failure cooldown is 10 seconds. If seeds accept then promptly receive an origin failure, `negcache.Lookup` has discarded that failure before this recheck (`negcache/negcache.go:175-184`). The recheck starts fresh origin downloads and interprets their `STARTED` replies as continuing progress.

This repeats within a cohort before activating backups, amplifying registry failures/rate limiting rather than observing terminal failures. Delegated failures have an additional gap because they intentionally do not populate the shared negative cache (`cmd/gantry/main.go:1864-1869`).

Please separate terminal-result observation from work admission, with appropriately scoped retention. Extending the shared negative cache alone does not solve delegated failures. Test STARTED, asynchronous failure, cooldown expiration, and resolver recheck; an immediately returned RecentlyFailed stub misses this lifecycle.

internal/gantry/mirror/mirror.go
```diff
@@ -1505,6 +1651,89 @@ func (s *Server) tryPeerFallback(ctx context.Context, w http.ResponseWriter, r *
...
+
+	logger.Debug("mirror: capacity response headers flushed", slog.Int64("size", size))
+
+	return true
+}
+
+func (s *Server) retryBusyFallback(ctx context.Context, w http.ResponseWriter, r *http.Request, d digest.Digest, kind ifaces.OriginRefKind, upstream, repo string, stream *livePeerStream, retryInterval, retryAfter time.Duration, logger *slog.Logger) peerFallbackResult {
+	for {
+		if stream != nil && stream.started {
+			if handled, result := s.serveStartedLocalHit(ctx, w, d, kind, upstream, repo, stream, logger); handled {
```
plombardi89: **[P2] Started-stream local serving is missing from ordinary rediscovery**

This helper handles local completion in the busy-retry loop, but ordinary rediscovery (`mirror.go:1615-1619`) and `recheckLocalAfterColdStart` (`1763-1768`) refuse to open local content once `stream.started`. The new pre-round-0 HEAD flush sets that flag before any cold pull.

If this node completes its selected cold pull while other seeds fail or are partitioned, only self may be discoverable. The mirror filters self and cannot consume its own available bytes, eventually closing an incomplete 200 response after rediscovery. No busy remote provider exists to enter this loop. A subsequent request may hit local cache, but the in-flight request unnecessarily fails.

Please use `serveStartedLocalHit` for ordinary rediscovery and post-cold-start checks too. The existing local-self-pull test (`mirror_coldstart_test.go:150-207`) omits `WithPeerRediscover`; enable that production combination and assert the same request completes locally without fetching self.

cchildress (approved):

internal/gantry/chairs/manager.go
vpatelsj: Fixed in 68c101fc.

Claim eligibility treated occupancy as proof of a live holder, so after a node-pool replacement every Lease still recorded a departed holder and no chair was ever free again. `attemptClaim` now also considers chairs unrenewed for five lease durations (`abandonedLeaseMultiple`), claiming them with `unresponsive=true`, which `Store.Claim` already gated on `Expired(now)`. Abandoned chairs are only used once no genuinely free chair remains, so a briefly slow holder keeps its seat.

`knownFull` was the other half of the trap: it meant "every chair records a holder" and short-circuited the claim path before reclamation could run. It now means "nothing is claimable" and accounts for abandoned chairs.

Tests in `internal/gantry/chairs/reclaim_test.go`: `TestManagerReclaimsAbandonedChairs` builds a full set of Leases with departed holders and older assignments and asserts a replacement node takes one. I confirmed it fails without the fix ("manager never reclaimed an abandoned chair"). `TestManagerDoesNotStealLiveChairs` is the counterweight and asserts a chair renewed a second ago is left alone.

cmd/gantry/chaircall.go
vpatelsj: Fixed in 68c101fc.

You are right that `ReadHeaderTimeout` does not bound body reads and `MaxBytesReader` caps size rather than time. The listener now sets `ReadTimeout` (15s), `WriteTimeout` (30s) and `IdleTimeout` (60s), and the handler is wrapped in a 256-slot admission bound that returns 503 rather than letting each request hold a goroutine and buffer.

Shutdown no longer blocks on a stuck handler: `stop` runs `Shutdown` under a 5s grace and falls back to `Close` when that expires, so `defer chairStop(context.Background())` cannot hold up process exit.

`cmd/gantry/chaircall_test.go` adds `TestChairServerBoundsStalledRequestBody`, which completes headers, declares a 4096-byte body, sends seven bytes and stalls. It passes in 15.01s, matching `chairReadTimeout`, which shows the new bound is what closes it. `TestChairServerShutdownIsBounded` covers the shutdown path.

internal/gantry/coord/chairhttp.go
vpatelsj: Fixed in 68c101fc.

`dialPinned` now derives its own `context.WithTimeout` covering both TCP establishment and the TLS handshake, and the `net.Dialer` carries the same timeout, so the dial no longer depends on a deadline that `Transport.getConn` never propagates. `MaxConnsPerHost` is also set, since the idle-pool limits did not bound connections still being established.

Test in `internal/gantry/coord/chairhttp_test.go`: `TestChairHTTPDialIsBoundedAgainstStalledTLS` points the client at a listener that accepts TCP and never negotiates TLS, and asserts the call returns bounded rather than hanging.

One note for the record: I could not confirm the `context.WithoutCancel` behaviour empirically in this pass, so I treated the premise as unproven. The mitigation is correct either way and costs nothing, so I applied it rather than arguing about it.

cmd/gantry/main.go
vpatelsj: Fixed in 68c101fc, and thank you for catching this. It was a regression I introduced earlier in the branch when I raised `maxUnavailable` from 8 to 50% for rollout speed; the arithmetic you describe is exactly right.

`Manager.Ready` now returns true when this node holds a chair, in addition to the existing `SelectableCount() >= SeedCount` condition. A chair holder is itself a usable seed, so gating it on cluster-wide occupancy was conflating cluster state with pod health. Generalising your case: the first batch is ceil(N/2), so every cluster under 16 nodes was affected, not only eight-node ones.

Test in `internal/gantry/chairs/ready_test.go`: `TestManagerReadyWithFewerHoldersThanSeedCount` asserts a single holder reports ready with far fewer than SeedCount chairs occupied.

You are also right that CI cannot see this: the kind e2e is a fresh install of exactly eight nodes, so it never exercises the base-to-head dependency.

internal/gantry/chairs/manager.go
vpatelsj: Fixed in 68c101fc.

`attemptClaim` returned before taking the snapshot, and `observe` is what retries `Connect`, so a non-holder that read 64 occupied chairs with all initial dials failed stopped observing entirely. The early return is now split: the "should I claim" conditions are computed into `skipClaim`, but the function only returns early when `bootstrapReady` is also true. A node that has not bootstrapped continues to snapshot and reconnect, and `skipClaim` is re-checked afterwards so claiming stays suppressed. A full snapshot now suppresses claims, not connection recovery, as you asked.

internal/gantry/mirror/mirror.go
vpatelsj: Fixed in 941a8e63.

`beginBusyResponse` now wraps the metadata HEAD in a 3s `busyMetadataTimeout` instead of inheriting the origin client's five-minute budget. A stalled registry costs the early header flush, which was already the behaviour on HEAD failure, and leaves the peer path intact rather than serialising it behind origin responsiveness.

internal/gantry/coldstart/chair.go
vpatelsj: Fixed in 941a8e63. This was the finding I considered most consequential, since it works against the whole point of the design.

`chairDispatchResult` now carries `stillPulling` separately from `accepted`. While any chair reports work already under way, the resolver keeps waiting and does not consume patience or recruit a backup cohort; the calling request bounds the wait. Escalation happens only when no chair reports work in flight, which still covers a silent cohort or one whose work ended without publishing.

Test in `internal/gantry/coldstart/patience_test.go`: `TestChairResolverWaitsOutHealthyLongPull` has every seed report ALREADY_PULLING indefinitely and asserts no chair outside the cohort is ever contacted. I confirmed it fails without the fix, recruiting chair 54.

internal/gantry/coldstart/chair.go
vpatelsj: Partially fixed in 941a8e63, and I want to be explicit about what is left.

Fixed: the recheck no longer reads STARTED as continuing progress. `PleasePullStarted` and `PleasePullAlreadyPulling` were both folded into "accepted"; they are now distinguished. A chair answering STARTED on a recheck had nothing in flight, so whatever it accepted earlier has ended without publishing, and that is now treated as a stall that consumes patience rather than as progress.

Not fixed: the recheck is still a work-starting `please_pull` rather than a status observation, so it can still admit fresh work. Properly separating terminal-result observation from work admission needs a status-only field on the wire, which I did not want to fold into this PR. I would rather leave that visible than claim it is done. Happy for you to reopen this thread if you would prefer it blocked on the wire change.

internal/gantry/mirror/mirror.go
vpatelsj: Fixed in 941a8e63.

Both call sites now use `serveStartedLocalHit` when the stream has started, instead of refusing to open local content. Ordinary rediscovery resumes from the local copy, and `recheckLocalAfterColdStart` does the same rather than returning false. As you note, the mirror filters self out of providers, so a node that completed its own cold pull was the only source and could not use it.

vpatelsj:
Thanks for the review — it was specific enough to act on directly, and two of the findings were real defects in code added on this branch. All nine are addressed across `68c101fc` and `941a8e63`, with per-finding detail in the threads.

| Finding | Fix |
|---|---|
| Obsolete occupied Leases | Chairs unrenewed for five lease durations are reclaimable; `knownFull` now means "nothing claimable" |
| Incomplete HTTPS bodies | Read/write/idle deadlines, 256-slot admission bound, bounded shutdown with forced close |
| Unbounded custom TLS dial | Deadline inside `dialPinned` covering TCP and handshake, plus `MaxConnsPerHost` |
| Eight-chair readiness deadlock | Holding a chair satisfies readiness on its own |
| Full snapshot suppresses bootstrap retry | Claim suppression separated from connection recovery |
| Slow origin HEAD | 3s bound on the metadata HEAD that precedes peer discovery |
| Healthy queued pulls duplicated | `stillPulling` tracked separately; escalation only when no chair reports work in flight |
| Failure recheck restarts work | STARTED no longer read as progress (see thread — partially addressed) |
| Started-stream local serving | `serveStartedLocalHit` used in rediscovery and post-cold-start |

Six new tests. I verified the two most consequential genuinely fail without their fix rather than passing vacuously: without the patience change the resolver recruits chair 54 outside the cohort, and without reclamation the manager never takes an abandoned chair. `TestManagerDoesNotStealLiveChairs` is the counterweight so reclamation cannot drift into stealing live chairs.

Two things worth calling out rather than burying:

**The readiness deadlock was mine.** I raised `maxUnavailable` from 8 to 50% earlier in this branch for rollout speed, which created exactly the condition described. Generalised, the first batch is ceil(N/2), so every cluster under 16 nodes was affected. Fixed in the readiness rule rather than by reverting, since the underlying constraint also broke any cluster smaller than SeedCount regardless of that setting.

**The failure-recheck thread is only partly addressed and I left it unresolved.** The interpretation is fixed — a recheck STARTED is no longer treated as progress — but the recheck is still a work-starting `please_pull` rather than a status observation. Fully separating observation from admission needs a status-only field on the wire, which I did not want to fold into this PR.

I also reconciled the design document as you asked: the rolling-interoperability section still promised Pod/Node RBAC and endpoint publication that this branch removes, and the cold-pull section described cohort retention that predated the escalation rule.

A 1,000-node benchmark is running now on the fixes. The previous four runs landed at 8.0007, 8.0007, 8.2007 and 8.1257 origin copies per layer against a design floor of 8. Worth noting the benchmark cannot refute the amplification findings either way: the origin is a private-endpoint ACR that returned zero failures, and those paths only engage under a slow or rate-limited registry.

cchildress (approved):

